PR #6, opened 2017-08-31
Typo error corrected in the pageviewregion.avsc

Kamal15:
- Comment to create the unneccessary 'PageViewsByUser' removed.
- Used try..resources to close the producer and consumer.

### DIFF
--- a/src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
+++ b/src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
@@ -112,8 +112,8 @@ public class PageViewRegionExampleDriver {
       consumer.subscribe(Collections.singleton(resultTopic));
       while (true) {
         final ConsumerRecords<String, Long> consumerRecords = consumer.poll(Long.MAX_VALUE);
-        for (ConsumerRecord<String, Long> record : consumerRecords) {
-          System.out.println(record.key() + ":" + record.value());
+        for (final ConsumerRecord<String, Long> consumerRecord : consumerRecords) {
+          System.out.println(consumerRecord.key() + ":" + consumerRecord.value());
         }
       }
     }
--- a/src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
+++ b/src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
@@ -18,6 +18,7 @@ import org.apache.avro.generic.GenericData.Record;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.avro.generic.GenericRecordBuilder;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
@@ -87,12 +88,12 @@ public class PageViewRegionExampleDriver {
         userProfileBuilder.set("region", regions[random.nextInt(regions.length)]);
         producer.send(new ProducerRecord<>(userProfilesTopic, user, userProfileBuilder.build()));
         // For each user generate some page views
-        IntStream.range(0, random.nextInt(10)).forEach(value -> {
+        for (int i=0; i<random.nextInt(10); i++) {
           pageViewBuilder.set("user", user);
           pageViewBuilder.set("page", "index.html");
           final Record record = pageViewBuilder.build();
           producer.send(new ProducerRecord<>(pageViewsTopic, null, record));
-        });
+        }
       }
     }
   }
@@ -111,7 +112,9 @@ public class PageViewRegionExampleDriver {
       consumer.subscribe(Collections.singleton(resultTopic));
       while (true) {
         final ConsumerRecords<String, Long> consumerRecords = consumer.poll(Long.MAX_VALUE);
-        consumerRecords.forEach(record -> System.out.println(record.key() + ":" + record.value()));
+        for (ConsumerRecord<String, Long> record : consumerRecords) {
+          System.out.println(record.key() + ":" + record.value());
+        }
       }
     }
   }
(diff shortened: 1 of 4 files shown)


src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
```diff
@@ -82,21 +81,20 @@ private static void produceInputs(String bootstrapServers, String schemaRegistry
...
-      // For each user generate some page views
-      IntStream.range(0, random.nextInt(10))
-        .mapToObj(value -> {
+    try (final KafkaProducer<String, GenericRecord> producer = new KafkaProducer<>(props)) {
+      for (final String user : users) {
+        userProfileBuilder.set("experience", "some");
+        userProfileBuilder.set("region", regions[random.nextInt(regions.length)]);
+        producer.send(new ProducerRecord<>(userProfilesTopic, user, userProfileBuilder.build()));
+        // For each user generate some page views
+        IntStream.range(0, random.nextInt(10)).forEach(value -> {
```
mjsax: I am not a Scale expert. Why do you remove `mapToObj` ?

src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
```diff
@@ -108,13 +106,12 @@ private static void consumeOutput(String bootstrapServers) {
...
-    consumer.subscribe(Collections.singleton(resultTopic));
-    while (true) {
-      final ConsumerRecords<String, Long> consumerRecords = consumer.poll(Long.MAX_VALUE);
-      for (final ConsumerRecord<String, Long> consumerRecord : consumerRecords) {
-        System.out.println(consumerRecord.key() + ":" + consumerRecord.value());
+    try (final KafkaConsumer<String, Long> consumer = new KafkaConsumer<>(consumerProperties)) {
+      consumer.subscribe(Collections.singleton(resultTopic));
+      while (true) {
+        final ConsumerRecords<String, Long> consumerRecords = consumer.poll(Long.MAX_VALUE);
+        consumerRecords.forEach(record -> System.out.println(record.key() + ":" + record.value()));
```
mjsax: We don't want to use Lambdas to be Java7 compliant. Cf. JavaDoc PageViewRegionExample.java#L43

\cc @miguno Or is it ok in the driver to change it?

src/main/resources/avro/io/confluent/examples/streams/pageviewregion.avsc
```diff
@@ -1,6 +1,6 @@
 {"namespace": "io.confluent.examples.streams.avro",
  "type": "record",
- "name": "PageViewStats",
+ "name": "PageViewRegion",
```
mjsax: Should it be `PageViewByRegion` ?

src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
Kamal15: @mjsax 
In the same example, `IntStream` is used which is Java8, so I have changed this one. No issues I'm happy to revert it back.

src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
Kamal15: For better readability, in this case, we don't need to do two operations.
`mapToObj` combined with `forEach` is same as the above one.

src/main/resources/avro/io/confluent/examples/streams/pageviewregion.avsc
Kamal15: I'm not sure.  The example was written in avro using `GenericRecord`, I have serialized it with plain Java serialization so found the error. I'm ok with either name.

src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
mjsax: Ack. Thanks for clarifying

src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
mjsax: @miguno Should the example be Java7 compliant or not?

I personally don't have a strong opinion. If it is Java8 already, it fine to update. Otherwise we need to fix `IntStream`, too... Let's wait for @miguno reply.

src/main/resources/avro/io/confluent/examples/streams/pageviewregion.avsc
mjsax: Don't have a strong opinion either. The output topic is called `PageViewByRegion` to it might be good to align names.

Kamal15:
@miguno 

call for review.

src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
miguno: I'd stick to Java 7 here.  There's an equivalent example `PageViewRegionLambdaExample` where we area using Java 8+.

src/main/resources/avro/io/confluent/examples/streams/pageviewregion.avsc
miguno: +1 to align names properly (= use `PageViewByRegion`).

Note that this should also require updating both the Java 7 and Java 8+ variants of our accompanying examples.

src/main/java/io/confluent/examples/streams/PageViewRegionExampleDriver.java
Kamal15: I have updated the `driver` class to use Java 7. 

Note that, many existing driver classes are still using `IntStream` which is introduced in Java 8. If ok to refactor, I'm ready to open a PR for them too.

miguno:
@mjsax @Kamal15 : Yes, ideally this PR should target the `3.3.0-post` branch.

Kamal15:
@miguno Opened a new PR. Pointing to the `3.3.0-post` branch. So, closing this one.
https://github.com/confluentinc/kafka-streams-examples/pull/11

miguno:
Thanks @Kamal15.  FYI: You can also modify/update the target branch of a PR via the GitHub web UI.

PR #9, opened 2017-09-08
Update example to KIP-120 changes

mjsax:


mjsax:
Sorry for import re-ordering. That is what Intellij does automatically... (I run with default setting, so not sure if anybody has different setting -- otherwise, this reordering should never happen... Any thoughts?)

src/main/java/io/confluent/examples/streams/interactivequeries/kafkamusic/KafkaMusicExample.java
```diff
@@ -178,19 +175,6 @@
   private static final String DEFAULT_BOOTSTRAP_SERVERS = "localhost:9092";
   private static final String DEFAULT_SCHEMA_REGISTRY_URL = "http://localhost:8081";
 
-  public static class CustomRocksDBConfig implements RocksDBConfigSetter {
```
guozhangwang: Cleanups are always welcome :)

PR #11, opened 2017-09-15
Typo error corrected in the pageviewregion.avsc

Kamal15:
- Comment to create the unneccessary 'PageViewsByUser' removed.
- Used try..resources to close the producer and consumer.

(cherry picked from commit a13f98c)

### DIFF
--- a/src/main/resources/avro/io/confluent/examples/streams/pageviewregion.avsc
+++ b/src/main/resources/avro/io/confluent/examples/streams/pageviewregion.avsc
@@ -1,6 +1,6 @@
 {"namespace": "io.confluent.examples.streams.avro",
  "type": "record",
- "name": "PageViewStats",
+ "name": "PageViewByRegion",
  "fields": [
      {"name": "user", "type": "string"},
      {"name": "page", "type": "string"},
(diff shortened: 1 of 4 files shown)


src/main/resources/avro/io/confluent/examples/streams/pageviewregion.avsc
```diff
@@ -1,6 +1,6 @@
 {"namespace": "io.confluent.examples.streams.avro",
  "type": "record",
- "name": "PageViewStats",
+ "name": "PageViewByRegion",
```
miguno: Just realized this while double-checking the new PR:  I think the "by region" part in "PageViewByRegion" is not correct because these records are simply (user, page, region) tuples -- they are not keyed by region.

I'd suggest to rename this to "PageViewWithRegion", as this schema is used as the output of a stream-table join that enriches page view events -- (user, page, industry, flags) tuples -- with region information.

Kamal15: Renamed it to `PageViewWithRegion`

miguno:
LGTM, thanks @Kamal15 !

miguno:
pint merge'd as well

PR #15, opened 2017-09-26
Fix dependency error (Jackson version got bumped in AK)

mjsax:


norwood:
you should change line 9 to use `rest-utils-parent`, then you can remove this version dependency and be automatically upgraded

norwood:
you can also remove a lot of the other deps in here, jetty, and everything that is using `${confluent.version}`

mjsax:
Updated this.

PR #16, opened 2017-10-02
add docker image tests + run them in CI

aayars:
These are the old tests from cp-docker-images, updated to run without docker-machine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,9 +14,10 @@
 # limitations under the License.
 
 # Builds a docker image for Confluent's example applications for the Kafka Streams API
-ARG DOCKER_REGISTRY
+ARG DOCKER_UPSTREAM_REGISTRY
+ARG DOCKER_UPSTREAM_TAG
 
-FROM ${DOCKER_REGISTRY}confluentinc/cp-base
+FROM ${DOCKER_UPSTREAM_REGISTRY}confluentinc/cp-base:${DOCKER_UPSTREAM_TAG}
 
 ARG STREAMS_VERSION
 ARG ARTIFACT_ID
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@
 
 # Builds a docker image for Confluent's example applications for the Kafka Streams API
 ARG DOCKER_UPSTREAM_REGISTRY
-ARG DOCKER_UPSTREAM_TAG
+ARG DOCKER_UPSTREAM_TAG=latest
 
 FROM ${DOCKER_UPSTREAM_REGISTRY}confluentinc/cp-base:${DOCKER_UPSTREAM_TAG}
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,13 @@
 #!/usr/bin/env groovy
 
 docker_oraclejdk8 {
+    dockerRepos = ['confluentinc/kafka-streams-examples']
+    dockerRegistry = '368821881613.dkr.ecr.us-west-2.amazonaws.com/'
+    dockerUpstreamRegistry = 'docker.io/'  // Temporary; use public images until new base images for trunk are published
+    dockerUpstreamTag = 'latest'  // Temporary; use trunk-latest when available
+    mvnPhase = 'integration-test'
+    mvnSkipDeploy = true
+    nodeLabel = 'docker-oraclejdk8-compose'
+    withPush = true
 
 }
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-git+https://github.com/confluentinc/confluent-docker-test-utils
+git+https://github.com/confluentinc/confluent-docker-utils@v0.0.5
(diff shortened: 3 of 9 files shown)


Dockerfile
```diff
@@ -14,9 +14,10 @@
 # limitations under the License.
 
 # Builds a docker image for Confluent's example applications for the Kafka Streams API
-ARG DOCKER_REGISTRY
+ARG DOCKER_UPSTREAM_REGISTRY
```
ewencp: should these have default values?

Jenkinsfile
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env groovy
 
 docker_oraclejdk8 {
+    dockerRepos = ['confluentinc/kafka-streams-examples']
+    dockerRegistry = '368821881613.dkr.ecr.us-west-2.amazonaws.com/'
```
ewencp: this should probably have some other dns name that we can redirect as we wish

requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/confluentinc/confluent-docker-test-utils
```
ewencp: this seems not ideal as it's not actually pinned to a version

setup.py
```diff
@@ -0,0 +1,29 @@
...
+
+from pip.req import parse_requirements
+import setuptools
+
+# Filters out relative/local requirements (i.e. ../lib/utils)
+remote_requirements = '\n'.join(str(r.req) for r in parse_requirements("requirements.txt", session='dummy') if r.req)
+
+setuptools.setup(
+    name='kafka-streams-examples-tests',
+    version='0.0.1',
```
ewencp: can we make this just match the version in the pom? in fact, can we just get the version from the pom?

setup.py
```diff
@@ -0,0 +1,29 @@
...
+
+    author="Confluent, Inc.",
+    author_email="partner-support@confluent.io",
```
ewencp: this seems like it's not the right email address for this repo, shouldn't this be to either the streams team or some more general address? i don't think there is anything partner related in this repo

setup.py
```diff
@@ -0,0 +1,29 @@
...
+
+    description='Docker image tests',
+
+    url="https://github.com/confluentinc/kafka-connect-images",
```
ewencp: wrong url for this repo?

test/fixtures/standalone-config.yml
```diff
@@ -0,0 +1,90 @@
...
+    environment:
+      ZOOKEEPER_CLIENT_PORT: 2181
+      ZOOKEEPER_TICK_TIME: 2000
+
+  kafka:
+    image: confluentinc/cp-kafka:latest
+    labels:
+    - io.confluent.docker.testing=true
+    environment:
+      KAFKA_BROKER_ID: 1
```
ewencp: not really related to this pr, but do we know if there is a reason we aren't using automatic broker ids?

test/test_kafka_streams_examples.py
```diff
@@ -0,0 +1,125 @@
...
+        # Verify access to the containerized application from inside its own container
+        self.is_kafka_music_app_healthy_for_service("kafka-streams-examples-bridge", port=17070)
+
+        # Verify outside access to the containerized application over the host network from a new container
+        list_running_app_instances_url_host = "http://{host}:{port}/kafka-music/instances".format(host="localhost", port=27070)
+        host_network_logs = utils.run_docker_command(
+            image="confluentinc/cp-kafka-streams-examples",
+            command=KAFKA_MUSIC_APP_HEALTH_CHECK.format(url=list_running_app_instances_url_host),
+            host_config={'NetworkMode': 'host'})
+        self.assertTrue("PASS" in host_network_logs)
```
ewencp: we seem to be mixing regular `assert`s and `self.assertTrue`

Dockerfile
aayars: If they're not given, the Docker default is to just use the latest from docker.io. We could provide defaults, but the implicit defaults are also okay.

Jenkinsfile
aayars: Unfortunately, this causes errors due to a principal name mismatch in the certificate :(

setup.py
aayars: The POM version is inherited from the parent -- we could just grab the parent version, but it's still pretty hacky.

I think we can just hard-code it to match the version for now.

Dockerfile
ewencp: huh, i wouldn't have thought including the `:` but not a tag following it would just use latest.

Jenkinsfile
ewencp: wow, how terrible and ugly. seems they've had requests for this for a couple of years now and still haven't implemented it.

miguno:
I have nothing to add to this PR review. :-)

Dockerfile
aayars: You're right about tag, added a default.

setup.py
aayars: `pint` should be able to bump the POM and setup.py in one go.

PR #17, opened 2017-10-04
change package name for SecurityProtocol from upstream change.

bbejeck:


bbejeck:
\cc @dguy @guozhangwang @mjsax 

Package name changed for `SecurityProtocol` from an upstream change in clients.

PR #18, opened 2017-10-05
Improving EventDeduplicationLambdaIntegrationTest

mjsax:
- some more cleanup for KIP-182

PR #19, opened 2017-10-11
Fix 4.0

mjsax:


mjsax:
This replaces #18 (fixed TODO and target against new `4.0.x` branch)

src/test/java/io/confluent/examples/streams/EventDeduplicationLambdaIntegrationTest.java
```diff
@@ -220,10 +222,19 @@ public void shouldRemoveDuplicatesFromTheInput() throws Exception {
     // on de-duplicating late-arriving records.
     long maintainDurationPerEventInMs = TimeUnit.MINUTES.toMillis(10);
 
+    // The number of segments has no impact on "correctness".
+    // Using more segments implies larger overhead but allows for more fined grained record expiration
+    // Note: the specified retention time is a _minimum_ time span and no strict upper time bound
```
bbejeck: nit: are the underscores intentional?

bbejeck:
Thanks @mjsax, one minor comment otherwise LGTM.

src/test/java/io/confluent/examples/streams/EventDeduplicationLambdaIntegrationTest.java
mjsax: Yes. To highlight.

src/main/scala/io/confluent/examples/streams/algebird/CMSStore.scala
```diff
@@ -256,7 +260,7 @@ class CMSStore[T: CMSHasher](override val name: String,
     */
   override def flush() {
     if (loggingEnabled) {
-      changeLogger.logChange(changelogKey, cms)
+      changeLogger.logChange(changelogKey, cms, timestampOfLastStateStoreUpdate)
```
guozhangwang: Hmm.. instead of passing the timestamp should we just get it from context as did in `StoreChangeLogger`?

mjsax: We cannot access the context here but get an exception. That's the reason for the change. Note, `flush()` is called on commit, ie, not during processing.

I am not sure why we were able to call `context.timestamp` in the old version -- some PR must have fixed this bug (not sure which one)

dguy: yep - i don't think this was previously called before i made the changes for KIP-182

PR #29, opened 2017-11-28
The incorrect jersey-container-servlet-core dependency forces servlet 2.x implementation that makes async request fail. 

magaton:
The current version of OrderService blows on POST request with 
javax.servlet.ServletException: java.lang.UnsupportedOperationException: Asynchronous processing not supported on Servlet 2.x container.
jersey-container-servlet needs to be a dependency since it is servlet 3.x implementation, not jersey-container-servlet-core which is servlet 2.x implementation. See https://mvnrepository.com/artifact/org.glassfish.jersey.containers/jersey-container-servlet-core and https://mvnrepository.com/artifact/org.glassfish.jersey.containers/jersey-container-servlet

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,7 +126,7 @@
         </dependency>
         <dependency>
             <groupId>org.glassfish.jersey.containers</groupId>
-            <artifactId>jersey-container-servlet-core</artifactId>
+            <artifactId>jersey-container-servlet</artifactId>
             <version>${jersey.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


guozhangwang:
cc @bbejeck @confluentinc/streams

bbejeck:
cc @benstopford

benstopford:
@magaton - how are you seeing this error? I'm just a little puzzled as the tests are still passing. Making the change to 3.x doesn't do any harm, but i'm intrigued?

magaton:
@benstopford ,
you don't see it in tests because the correct dependency (jersey-container-servlet ) is pulled in the test scope.
The problem occurs when you execute the main method. java -jar 
Then the jersey-container-servlet-core is pulled.

benstopford:
Ah, that make sense. Thanks for the PR @magaton

PR #32, opened 2017-11-29
boolean running is never set to true

magaton:
typo - incorrect method call in main()


benstopford:
Looks good. Thanks @magaton.

PR #37, opened 2017-12-05
Fix word count to query other instances

bbejeck:


bbejeck:
@dguy @guozhangwang @mjsax @miguno 

Updated the example to query across instances for individual word counts

guozhangwang:
@bbejeck Since the user reported seeing this issue in 3.3.0 also, should we submit the PR against 3.3.0-post and merge all the way down to other branches?

bbejeck:
Yes.  I overlooked the fact that is was in 3.3.0-post as well.

mjsax:
@bbejeck No need to close and reopen a new PR -- you can also change the target branch for the PR. Just FYI.

PR #38, opened 2017-12-06
Fix word count to query other instances

bbejeck:


bbejeck:
@guozhangwang @mjsax @dguy @miguno

PR #41, opened 2017-12-20
KAFKA-1166: fix flaky EndToEnd integration test

mjsax:
- increase RS timeout (for two tests)
 - some code cleanup


bbejeck:
Thanks, @mjsax LGTM.  FWIW I ran `EndToEndTest` 50 times, and they all passed.

PR #48, opened 2018-01-04
Fix findbugs errors and incorrect indexing in args

ewencp:


src/main/java/io/confluent/examples/streams/GlobalKTablesExampleDriver.java
```diff
@@ -91,6 +91,7 @@ private static void receiveEnrichedOrders(final String bootstrapServers,
     while(received < expected) {
       final ConsumerRecords<Long, EnrichedOrder> records = consumer.poll(Long.MAX_VALUE);
       records.forEach(record -> System.out.println(record.value()));
+      received += records.count();
```
mjsax: Seems this was broken from the beginning on... *sigh*

Could you do me a favor and fix the indention of the parameter of this (an maybe all other) method in this class, too?

ewencp: I fixed them. tbh this repo seems like it could use quite a bit of love... seems like code coverage is pretty low, it seems surprising that the indexing of args could be broken the way it was if the apps had proper testing, and the code style is quite a mess (checkstyle seems to have been disabled for some reason.

mjsax: `main()` is not used in tests... it's just there to run the example "stand alone"

Agreed that we should enable checkstyle if it is really disabled. \cc @confluentinc/streams

ewencp: i mean, i get why its there but the code in there is still testable. maybe needs refactoring if you don't want to call main() directly.

for checkstyle, it is disabled, see https://github.com/confluentinc/kafka-streams-examples/blob/4.0.0-post/pom.xml#L421. based on the blame, i'm guessing this was due to migrating to common pom but not wanting to have to deal with all that fallout

PR #54, opened 2018-01-17
MINOR: increase SR timeout to stabilize EndToEndTest

mjsax:


mjsax:
If this does not work, we need to consider rewriting the test. SR folk suggested to increase to 60 seconds.

PR #56, opened 2018-01-18
HOTFIX: tests with method ambiguity

bbejeck:
3 more tests needing method ambiguity fixes


bbejeck:
Found 3 more tests needing the fix for KIP-149

bbejeck:
already addressed in another PR closing

PR #59, opened 2018-01-19
MINOR: fix race condition for Schema-Registry startup

mjsax:


src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
```diff
@@ -76,11 +78,25 @@ public void start() throws Exception {
...
+    int retries = 5;
+    while (retries > 0) {
+      schemaRegistry = new RestApp(
+          InstanceSpec.getRandomPort(),
+          zookeeperConnect(),
+          KAFKA_SCHEMAS_TOPIC, AVRO_COMPATIBILITY_TYPE);
+      try {
+        schemaRegistry.start();
+        break;
+      } catch (final BindException exception) {
```
guozhangwang: We once had a similar issue for the port-in-use root cause in AK: https://issues.apache.org/jira/browse/KAFKA-1501

Instead of retrying, I think a better approach was done there, is to "allow the port to be 0 to make the kernel give us a random port": https://reviews.apache.org/r/31806/diff/3#1

src/main/java/io/confluent/examples/streams/TopArticlesLambdaExample.java
```diff
@@ -142,7 +142,6 @@ static KafkaStreams buildTopArticlesStream(final String bootstrapServers,
     // Specify default (de)serializers for record keys and for record values.
     streamsConfiguration.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
     streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, GenericAvroSerde.class);
-    streamsConfiguration.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 1000);
```
bbejeck: more out of curiosity, why this change?

bbejeck:
Just one question, but otherwise LGTM

src/main/java/io/confluent/examples/streams/TopArticlesLambdaExample.java
mjsax: Because commit interval is set to 10 seconds in line 149.

PR #60, opened 2018-01-19
KSTREAM-1264: fix unstable EndToEndTest

mjsax:


mjsax:
The method `randomFreeLocalPort()` is subject to a race condition as it does not "block" the port; thus, it could happen that the port gets occupied by another process after the method returned. Need to retry.

mjsax:
Call for review @confluentinc/streams

benstopford:
Looks good

benstopford:
Thank you @mjsax :)

PR #71, opened 2018-02-14
MINOR: enable DEBUG logging for all tests by default

mjsax:


mjsax:
I think we should enable DEBUG logging for test by default. Often if a test fails, we have very limited information why.

src/test/resources/log4j.properties
```diff
@@ -0,0 +1,5 @@
+log4j.rootLogger=DEBUG, stdout
```
jaceklaskowski: Shouldn't `stdout` have the corresponding `log4j.appender.stdout.target=System.out`?

mjsax: I guess the name is just a little confusion... we want to capture debug-level logs, not what's written to System.out and print them to the console/stdout...

Or do I miss understand your comment?

jaceklaskowski: I'm just referring to the name as is that has no corresponding `appender` defined whether it's `System.out` or not. I thought it's required to have a `log4j.appender.[name].target` entry defined.

mjsax: Default target is `System.out` -- no need to specify.

PR #72, opened 2018-02-15
MINOR: Fix incorrect OrderServiceTest

mjsax:
OrderServiceTest need to cleanup the used order-topic between tests. Otherwise, the test does not start with a clean test environment.

The test sometimes fails with
```
org.junit.ComparisonFailure: expected:<...customerId=2, state=[CRE]ATED, product=JUMPER...> but was:<...customerId=2, state=[VALID]ATED, product=JUMPER...>
	at io.confluent.examples.streams.microservices.OrdersServiceTest.shouldPostOrderAndGetItBack(OrdersServiceTest.java:79)
```

Updating the customer-id in each test case shows, that data from another test run is consumed:
```
org.junit.ComparisonFailure: expected:<...{id='1', customerId=[4], state=VALIDATED, p...> but was:<...{id='1', customerId=[3], state=VALIDATED, p...>
	at io.confluent.examples.streams.microservices.OrdersServiceTest.shouldGetOrderByIdWhenOnDifferentHost(OrdersServiceTest.java:175)
```
This issue was not exposed before, as all test used the same client-id -- I suspect that the test only passed because all tests used the same Order-object and thus, even if reading data from a wrong test test, this issue was not detected as the test saw the data as expected.

Not sure, why the test sometime passed -- must be some race condition between test runs -- maybe test execution order.

### DIFF
--- a/src/test/java/io/confluent/examples/streams/microservices/OrdersServiceTest.java
+++ b/src/test/java/io/confluent/examples/streams/microservices/OrdersServiceTest.java
@@ -9,6 +9,7 @@ import io.confluent.examples.streams.microservices.domain.beans.OrderBean;
 import io.confluent.examples.streams.microservices.util.MicroserviceTestUtils;
 import io.confluent.examples.streams.microservices.util.Paths;
 import org.junit.After;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
@@ -19,11 +20,11 @@ import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.Response;
 import java.net.HttpURLConnection;
+import java.util.Collections;
 
 import static io.confluent.examples.streams.avro.microservices.Order.newBuilder;
 import static io.confluent.examples.streams.microservices.domain.beans.OrderId.id;
 import static io.confluent.examples.streams.microservices.util.MicroserviceUtils.MIN;
-import static java.util.Arrays.asList;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON_TYPE;
 import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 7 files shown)


mjsax:
Going to do multiple test runs to make sure it's stable...

mjsax:
Test failed again with similar error... not sure why -- seems this is not the right fix...

mjsax:
OrderService is a stateful Kafka Streams app. I think, that local state on disk can lead to old data, too. Added one more commit to clean local state between test runs. Let's see if this fixed the flakiness.

src/test/java/io/confluent/examples/streams/microservices/OrdersServiceTest.java
```diff
@@ -35,17 +43,36 @@
...
+      rest2.cleanLocalState();
+    }
+  }
+
+  @Before
+  public void prepareKafkaCluster() throws Exception {
+    CLUSTER.deleteTopicsAndWait(30000, Topics.ORDERS.name());
+    CLUSTER.createTopic(Topics.ORDERS.name());
+    Schemas.configureSerdesWithSchemaRegistryUrl(CLUSTER.schemaRegistryUrl());
+    // TODO remove debugging code from below
```
guozhangwang: nit: Is this `TODO` valid?

src/test/java/io/confluent/examples/streams/microservices/util/MicroserviceTestUtils.java
```diff
@@ -182,46 +184,40 @@ public static Properties propsWith(KeyValue... props) {
     return properties;
   }
 
-  public static <K, V> void send(Topic topic, KeyValue<K, V>... stuff) {
-    KafkaProducer<Long, Order> producer = new KafkaProducer(producerConfig(CLUSTER),
-        topic.keySerde().serializer(), topic.valueSerde().serializer());
-    try {
+  public static <K, V> void send(Topic<K, V> topic, Collection<KeyValue<K, V>> stuff) {
+    try(KafkaProducer<K, V> producer = new KafkaProducer<>(producerConfig(CLUSTER),
```
guozhangwang: nit: space?

mjsax:
Going to trigger couple of rebuilds to make sure it's stable...

mjsax:
Run 10 failed unrelated:
```
java.lang.AssertionError: 

Expected: <{}>
     but: was <{donna=3, becca=2, john=1}>
	at io.confluent.examples.streams.WikipediaFeedAvroExampleTest.shouldRunTheWikipediaFeedExample(WikipediaFeedAvroExampleTest.java:129)
```

@bbejeck Wanna add this to the list :D

PR #76, opened 2018-02-22
Update zk timeout to 60 seconds Fix test assertion order

bbejeck:


bbejeck:
This PR will address some of the failures due to ZK connection timeouts

bbejeck:
merged with pint

PR #80, opened 2018-02-26
MINOR: add pluginRepo

mjsax:


mjsax:
PluginRepo was removed from 3.3.1-post but added back to 4.0.0-post later without back porting

PR #81, opened 2018-03-02
Currected Scripts name

abdheshkumar:


guozhangwang:
@abdheshkumar Thanks for the PR. The confluent examples are based on the confluent open source release which uses its own shell scripts, and it does not have the `.sh` suffix.

mjsax:
What is the status of this PR? Should it be closed?

miguno:
Yes, this can be closed.  "Won't Fix" -- works as intended. :)

PR #83, opened 2018-03-13
MINOR: fix example usage of WindowedSerdes

vvcephei:


src/main/java/io/confluent/examples/streams/TopArticlesLambdaExample.java
```diff
@@ -50,7 +50,7 @@
  *
  * In this example, we count the TopN articles from a stream of page views (aka clickstreams) that
  * reads from a topic named "PageViews". We filter the PageViews stream so that we only consider
- * pages of type article, and then map the recored key to effectively nullify the user, such that
+ * pages of type article, and then map the recorded key to effectively nullify the user, such that
```
mjsax: I think "record key" was correct -- the key of the record.

vvcephei: oh, shoot. I'll fix it.

vvcephei:
@mjsax I fixed that typo.

Also, I increased the timeout of TopArticlesLambdaExampleTest to 60s. We had one build that filed because it hit the old timeout of 30s. The subsequent build passed after 25s, so I think 30s is just too close.

mjsax:
Thanks @vvcephei ! Sounds good!

PR #86, opened 2018-03-19

bbejeck:
`rest-utils` removed  the `jackson.verson` property to rely on the one specified in common pom to use the same version across the platform.  With these changes all tests pass.


bbejeck:
Should this go against `3.3.0-post` instead?  I targeted this branch as it was the same branch `rest-utils` made the changes.

mjsax:
3.3.0-post sound better to me.

bbejeck:
that's what I was thinking, just wanted confirmation thanks.

PR #87, opened 2018-03-19

bbejeck:


bbejeck:
sorry for the mess, but merging against 3.3.0-post will break so it needs to go against 3.3.x

bbejeck:
Asking reviews to post again

PR #88, opened 2018-03-19

bbejeck:


bbejeck:
Remove `jackson.version` setting and use setting from common.

bbejeck:
sorry for the confusing mess, please re-vote for the record

PR #89, opened 2018-03-20
Compilation failure of all the classes which are dependent on "Avro generated sources" due to "Avro generated sources" missing in the build path

ajithprasadanand:
The mvn compiled classes were already present in generated-sources but the IDE was not picking it. I had to add the following line in the pom:

**${project.build.directory}/generated-sources**

```
<plugin>
                <groupId>org.codehaus.mojo</groupId>
                <artifactId>build-helper-maven-plugin</artifactId>
                <version>1.10</version>
                <executions>
                    <execution>
                        <id>add-source</id>
                        <phase>generate-sources</phase>
                        <goals>
                            <goal>add-source</goal>
                        </goals>
                        <configuration>
                            <sources>
                                <source>src/main/scala</source>
				<source>${project.build.directory}/generated-sources</source>
                            </sources>
                        </configuration>
                    </execution>
```


ajithprasadanand:
[clabot:check]

PR #96, opened 2018-03-23
MINOR: adding schema registry configs

bbejeck:
Added configs for Schema Registry in an effort to see if this helps with examples tests flakiness since most of the failure involve Schema Registry

- Increased store init timeout to 90 seconds
- Increased store operation timeout from 500ms to 10s
- Set debug flag to true


bbejeck:
merged with pint

PR #98, opened 2018-03-26
Modified GlobalKTablesExample and its driver to show correct consumption rate

badaiaqrandista:
Modified GlobalKTablesExample and its driver to show correct consumption rate.

This is my first pull request. Any input is appreciated.


badaiaqrandista:
Workaround should not be in example.

PR #103, opened 2018-04-05
MINOR: take out fix until 3.3.2 is released

bbejeck:
Taking out changes for setting `RestApp` properties until 3.3.2 is released.


bbejeck:
merged with pint

PR #119, opened 2018-05-23
MINOR: Bump version javax.ws.rs

bbejeck:
rest-utils upgraded to latest versions of jetty and jersey, needed to update version of javax.ws.rs-api


bbejeck:
@mjsax looked into that but it looks like the parent pom does not have `javax.ws.rs-api` as a dependency but I'll double check

bbejeck:
waiting for jenkins to finish before merging

PR #120, opened 2018-05-27
Fix examples for RecordHeader merge

bbejeck:


bbejeck:
Needed to update the calls to `RecordCollector` to take an extra parameter for `RecordHeaders`

src/main/scala/io/confluent/examples/streams/algebird/CMSStoreChangeLogger.scala
```diff
@@ -44,7 +44,7 @@ class CMSStoreChangeLogger[K, V](val storeName: String,
     if (collector != null) {
       val keySerializer = serialization.keySerializer
       val valueSerializer = serialization.valueSerializer
-      collector.send(this.topic, key, value, this.partition, timestamp, keySerializer, valueSerializer)
+      collector.send(this.topic, key, value, null, this.partition, timestamp, keySerializer, valueSerializer)
```
bbejeck: using `null` here as `context.headers()` results in an `IllegalStateException` as the `RecordContext` is null since the key and value is passed to the `RecordCollector` here on store flush.

bbejeck: using `null` here as `context.headers()` results in an `IllegalStateException` as the `RecordContext` is null since the key and value is passed to the `RecordCollector` here on store flush.

mjsax: Don't understand the comment... If passing `null` result in `IllegalStateException` why is the change a proper fix?

bbejeck:
I need to pass in null, calling context.headers() causes the
IllegalStateException

src/main/scala/io/confluent/examples/streams/algebird/CMSStoreChangeLogger.scala
bbejeck: Sorry if my comments aren't clear. 
 
We need to pass `null` for headers as calling `context.headers()` is what causes the `IllegalStateException`.

Passing null for `headers` follows the pattern in  `org.apache.kafka.streams.state.internals.StoreChangeLogger`

mjsax: Ah. Makes sense know. Thanks for clarification.

PR #131, opened 2018-07-18
KSTREAMS-1609: Remove refs to --new-consumer

bbejeck:
Remove refs to `--new-consumer` otherwise users get an error of `new-consumer is not a recognized option`.


bbejeck:
merged with pint

PR #138, opened 2018-07-24
Serdes cached in static variable

benstopford:
Serdes must be initialised with the schema registry but are being cached in a static Joined variable before the schema registry is configured.


ybyzek:
LGTM

PR #140, opened 2018-08-05
KSTREAM-1647: Fix KafkaMusicExampleTest

mjsax:
Fix race condition by waiting for expected output in stores

### DIFF
--- a/src/test/java/io/confluent/examples/streams/interactivequeries/kafkamusic/KafkaMusicExampleTest.java
+++ b/src/test/java/io/confluent/examples/streams/interactivequeries/kafkamusic/KafkaMusicExampleTest.java
@@ -226,13 +226,11 @@ public class KafkaMusicExampleTest {
 
     // Wait until the all-songs state store has some data in it
     TestUtils.waitForCondition(() -> {
-      final ReadOnlyKeyValueStore<String, KafkaMusicExample.TopFiveSongs> topFiveStore;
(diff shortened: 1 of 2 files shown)


mjsax:
@confluentinc/streams It's unclear to me, why this test started to fail now. Must be some change in AK `trunk` but not sure which one it could be. Anyway, the previous wait conditions was incorrect anyway. I am just wondering why it did pass before (would expect that it should have hit the race before, too).

src/test/java/io/confluent/examples/streams/interactivequeries/kafkamusic/KafkaMusicExampleTest.java
```diff
@@ -225,12 +226,44 @@ public void shouldCreateChartsAndAccessThemViaInteractiveQueries() throws Except
 
     // Wait until the all-songs state store has some data in it
     TestUtils.waitForCondition(() -> {
-      final ReadOnlyKeyValueStore<Long, Song>
-          songsStore;
+      final ReadOnlyKeyValueStore<String, KafkaMusicExample.TopFiveSongs> topFiveStore;
+      final ReadOnlyKeyValueStore<String, KafkaMusicExample.TopFiveSongs> topFiveByGenreStore;
```
vvcephei: It looks like these aren't used outside of the try/catch, so I think you can declare them at initialization.

mjsax:
Updated this. Will merge after Jenkins passed.

Issue #150, opened 2018-08-30
Microservices: ValidationsAggregatorService.java failure (DEVX-244)


ybyzek:
https://github.com/confluentinc/kafka-streams-examples/blob/5.0.x/src/main/java/io/confluent/examples/streams/microservices/ValidationsAggregatorService.java#L41-L54

Problem: fails for the same reason that `EmailService` failed (and was fixed in PR #148, #138)

Serdes must be initialised with the schema registry but are being cached in a static Joined variable before the schema registry is configured.

mjsax:
Closing. PR was merged.

PR #151, opened 2018-08-30
DEVX-244: fix serdes in ValidationsAggregatorService

ybyzek:
Issue #150

### DIFF
--- a/src/main/java/io/confluent/examples/streams/microservices/ValidationsAggregatorService.java
+++ b/src/main/java/io/confluent/examples/streams/microservices/ValidationsAggregatorService.java
@@ -38,6 +38,20 @@ public class ValidationsAggregatorService implements Service {
 
   private static final Logger log = LoggerFactory.getLogger(ValidationsAggregatorService.class);
   private static final String ORDERS_SERVICE_APP_ID = "orders-service";
+  private final Consumed<String, OrderValidation> serdes1 = Consumed
+      .with(ORDER_VALIDATIONS.keySerde(), ORDER_VALIDATIONS.valueSerde());
+  private final Consumed<String, Order> serdes2 = Consumed.with(ORDERS.keySerde(),
+      ORDERS.valueSerde());
+  private final Serialized<String, OrderValidation> serdes3 = Serialized
+      .with(ORDER_VALIDATIONS.keySerde(), ORDER_VALIDATIONS.valueSerde());
+  private final Joined<String, Long, Order> serdes4 = Joined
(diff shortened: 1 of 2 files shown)


mjsax:
Cannot follow (I did read the ticket and the other PRs): what is the issue with those `Serdes` being `private final`? Can you elaborate?

vvcephei:
@mjsax @ybyzek Just for curiosity's sake, what was the reason? I also couldn't find it in the ticket.

ybyzek:
@vvcephei , @benstopford gave a much clearer explanation in #138 .  (I may have mis-spoken about the `final` keyword)

vvcephei:
Yikes! That's a tricky dependency.

I see, so the actual problem was just that the serdes were created during class construction, when they actually must be preceeded by a call to `io.confluent.examples.streams.microservices.util.MicroserviceUtils#parseArgsAndConfigure`

Then it should be ok to make them `private final` local variables.

Another way to accomplish this would be to change only the `main` method:
```java
  public static void main(final String[] args) throws Exception {
    final String bootstrapServers = parseArgsAndConfigure(args);
    final ValidationsAggregatorService service = new ValidationsAggregatorService();
    service.start(bootstrapServers, "/tmp/kafka-streams");
    addShutdownHookAndBlock(service);
  }
```

It should also be possible to set a flag in `Schemas` to ensure that no one can get the topics until after the call to `io.confluent.examples.streams.microservices.domain.Schemas#configureSerdesWithSchemaRegistryUrl`

ybyzek:
@vvcephei -- just to be clear, are you ok with the PR as it is right now?  Or is there a specific recommendation you are making?

benstopford:
Both approaches should be fine, although @vvcephei's suggestions seem clearer. LGTM

src/main/java/io/confluent/examples/streams/microservices/ValidationsAggregatorService.java
```diff
@@ -68,6 +54,21 @@ private KafkaStreams aggregateOrderValidations(
       final String stateDir) {
     final int numberOfRules = 3; //TODO put into a KTable to make dynamically configurable
 
+    Consumed<String, OrderValidation> serdes1 = Consumed
```
bbejeck: needs to be `final` here and below to `serdes7`

ybyzek:
I have recoded to use @vvcephei 's suggestion.  Please review updated code changes (much cleaner).

vvcephei:
Thanks, @ybyzek ! LGTM.

(sorry, I didn't see your earlier question until now)

PR #152, opened 2018-08-30
MINOR: Make serdes final

bbejeck:
Missed this during review, `serdes` variable should be `final`.


bbejeck:
waiting for build to finish

bbejeck:
merged with pint

PR #135, opened 2018-07-23
KSTREAMS-1621: Initial integration of monitoring interceptors

bbejeck:
Example for making implementing `MonitoringInterceptors` easier when working with CE and C3

PR #153, opened 2018-08-31
KSTREAMS-1712: Add option for configuring Producers, Consumers with interceptors

bbejeck:
Refactored `MonitoringInterceptorUtils` to be able to configure individual producers and consumers in addition to configuring a streams application with Confluent Monitoring interceptors.

For testing, I picked two tests and added configuration for monitoring interceptors and confirmed configuration and usage via logs.

### DIFF
--- a/src/main/java/io/confluent/examples/streams/utils/MonitoringInterceptorUtils.java
+++ b/src/main/java/io/confluent/examples/streams/utils/MonitoringInterceptorUtils.java
@@ -5,6 +5,8 @@ import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.streams.StreamsConfig;
 
 import java.util.Properties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Utility helper class that will enable Monitoring Interceptors when
@@ -18,7 +20,7 @@ public class MonitoringInterceptorUtils {
 
     private static final String CONSUMER_INTERCEPTOR = "io.confluent.monitoring.clients.interceptor.MonitoringConsumerInterceptor";
     private static final String PRODUCER_INTERCEPTOR = "io.confluent.monitoring.clients.interceptor.MonitoringProducerInterceptor";
-
+    private static final Logger LOG = LoggerFactory.getLogger(MonitoringInterceptorUtils.class);
 
     private MonitoringInterceptorUtils() {
     }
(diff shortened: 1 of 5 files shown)


src/main/java/io/confluent/examples/streams/utils/MonitoringInterceptorUtils.java
```diff
@@ -23,22 +23,45 @@
...
     }
 
-    public static void maybeConfigureInterceptors(final Properties streamsConfig) {
-        try {
-            Class.forName(PRODUCER_INTERCEPTOR);
-        } catch (final ClassNotFoundException e) {
-            System.out.println("Monitoring Producer Interceptors not found, skipping");
-            return;
+    public static void maybeConfigureInterceptorsStreams(final Properties streamsConfig) {
+        if (hasMonitoringConsumerInterceptor() && hasMonitoringProducerInterceptor()) {
```
vvcephei: should this be two independent `if` blocks so that we can add the consumer or producer interceptors independently?

Or is that not a thing?

src/main/java/io/confluent/examples/streams/utils/MonitoringInterceptorUtils.java
```diff
@@ -23,22 +23,45 @@
...
+    private static boolean hasMonitoringInterceptor(final String className) {
+        boolean hasInterceptor = true;
         try {
-            Class.forName(CONSUMER_INTERCEPTOR);
+            Class.forName(className);
         } catch (final ClassNotFoundException e) {
-            System.out.println("Monitoring Consumer Interceptors not found, skipping");
-            return;
+            final String interceptorTypeShortName = className.substring(className.lastIndexOf('.'));
+            System.out.println(String.format("%s not found, skipping", interceptorTypeShortName));
```
vvcephei: Should this be a log?

src/main/java/io/confluent/examples/streams/utils/MonitoringInterceptorUtils.java
bbejeck: From previous comments https://github.com/confluentinc/kafka-streams-examples/pull/135#discussion_r204580802 , I'll leave as is for now and we can change later if we want.

src/main/java/io/confluent/examples/streams/utils/MonitoringInterceptorUtils.java
bbejeck: ack

bbejeck:
waiting for build to finish, then merging

bbejeck:
merged with pint

guozhangwang:
Thanks @bbejeck !!

PR #171, opened 2018-10-02
 ST-1153: Switch to use cp-base-new and bash-config to hide passwords

maxzheng:
Passwords were printed to stdout, and this will hide them by default with an option to turn on trace.
As for cp-base naming, Alex renamed cp-base to cp-base-new in common-docker to avoid confusion with cp-base from cp-docker-images.

See related change confluentinc/common-docker#18


test/test_kafka_streams_examples.py
```diff
@@ -49,7 +49,9 @@ def test_default_config(self):
         #       we should also validate some REST API responses (see `test_kafka_rest.py`)?
 
     def test_secure_config(self):
-        output = self.cluster.run_command_on_service("secure-config", "bash -c 'while [ ! -f /tmp/config-is-done ]; do echo waiting && sleep 1; done; echo PASS'")
+        output = self.cluster.run_command_on_service(
+            "secure-config",
```
mohnishbasha: Can also add the line break as self.cluster.run_command_on_service("secure-config",

maxzheng: The 2nd line is still a bit long / either way is good style if it works.

include/etc/confluent/docker/configure
```diff
@@ -14,10 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -o nounset \
-    -o errexit \
-    -o verbose \
-    -o xtrace
+. /etc/confluent/docker/bash-config
```
mohnishbasha: Is this file being made available from base?

maxzheng: yep

Issue #175, opened 2018-10-12
Access to `--reset` parameter within ApplicationResetExample needs fix


1123:
`args[0]` is used to pass the bootstrap servers address. 
`args[1]` should be used for passing the --reset flag. 

Currently the example code expects both parameters in `args[0]`, which does not work.

PR #176, opened 2018-10-12
ApplicationResetExample: bugfix for the --reset flag

1123:
See https://github.com/confluentinc/kafka-streams-examples/issues/175

### DIFF
--- a/src/main/java/io/confluent/examples/streams/ApplicationResetExample.java
+++ b/src/main/java/io/confluent/examples/streams/ApplicationResetExample.java
@@ -113,7 +113,7 @@ import java.util.Properties;
  * Thus, restart the application via:
  * <pre>
  * {@code
- * $ java -cp target/kafka-streams-examples-5.0.0-SNAPSHOT-standalone.jar io.confluent.examples.streams.ApplicationResetExample --reset
+ * $ java -cp target/kafka-streams-examples-5.0.0-SNAPSHOT-standalone.jar io.confluent.examples.streams.ApplicationResetExample localhost:9092 --reset
  * }</pre>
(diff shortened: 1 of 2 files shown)


1123:
[clabot:check]

src/main/java/io/confluent/examples/streams/ApplicationResetExample.java
```diff
@@ -161,7 +161,7 @@ public static KafkaStreams run(final String[] args, final Properties streamsConf
     final KafkaStreams streams = new KafkaStreams(builder.build(), streamsConfiguration);
 
     // Delete the application's local state on reset
-    if (args.length > 0 && args[0].equals("--reset")) {
+    if (args.length > 1 && args[1].equals("--reset")) {
```
mjsax: I think it would be better to make the example more flexible. IMHO, this method could just accept a `boolean doReset` instead of `String[] args` and we should add code in `main()` method, to check for 3 cases (zero, one, two arguments passed) and calls `run()` accordingly.

1123: indeed. Would be much better. I will update the PR.

1123: It might also be better to parse options with something like commons-cli. Currently the first argument is expected to be bootstrap-servers, and the second one the reset option. With named arguments the order would matter.

mjsax: We could. IMHO, it's ok for this example to keep it simple.

mjsax:
Thanks a lot for the PR!

PR #181, opened 2018-10-29
Use scala-library 2.12.7 and replace assert4J with scalatest assert

ijuma:
We upgraded to Scala 2.12 in the common pom and we had to upgrade the
scalatest dependency to get a build for that Scala version.

The assert4J change is due to an overload ambiguity when using Scala 2.12.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
         <java.version>1.8</java.version>
         <licenses.version>5.2.0-SNAPSHOT</licenses.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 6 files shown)


ijuma:
The assertThat version doesn't compile so...

bbejeck:
>The assertThat version doesn't compile so...

I'm fine with it as is, I wasn't aware of the dropped support for `assertThat`, more just making a general comment which probably wasn't necessary on my part.

ijuma:
@bbejeck I updated the PR description to include more information (it was only in the commit I added after the original PR previously).

bbejeck:
thanks @ijuma!

Issue #189, opened 2019-01-11
Microservices example: "sending an email to" does not print info


ybyzek:
https://github.com/confluentinc/kafka-streams-examples/blob/5.1.0-post/src/main/java/io/confluent/examples/streams/microservices/EmailService.java#L96-L97

Results in:

```bash
[2019-01-11 09:49:21,112] WARN [EmailService-771c6cc9-5056-4faf-83f7-dcd0578f03d9-StreamThread-1] Sending an email to: 
Customer:%s
Order:%s
Payment%s (io.confluent.examples.streams.microservices.EmailService)
```

The formatting is not properly printing the values.

Code change should be either one of these two options:

```java
log.warn("Sending an email to: \nCustomer: " + details.customer + "\nOrder: " + details.order + "\nPayment: " + details.payment);
log.warn("Sending an email to: \nCustomer:{}\nOrder:{}\nPayment{}", details.customer, details.order, details.payment);
```

ybyzek:
Being fixed as part of #190

PR #190, opened 2019-01-11
DEVX-544: dynamic routing example

ybyzek:
The goal of this PR is to show dynamic routing using TopicNameExtractor

This PR also fixes https://github.com/confluentinc/kafka-streams-examples/issues/189


mjsax:
@ybyzek Should this target 5.0.0-post? Dynamic routing was added in 2.0/5.0 release.

ybyzek:
> @ybyzek Should this target 5.0.0-post? Dynamic routing was added in 2.0/5.0 release.

@mjsax in theory it could have, but it needs to stay in sync with the corresponding changes going into [another repo with this PR](https://github.com/confluentinc/examples/pull/102) .  As long as we have an example to point to, even going into 5.1.0-post, I think it's ok.  WDYT?

mjsax:
Ack. Fine with me to only have it for 5.1 :)

miguno:
@ybyzek : Thanks so much.  Could you please also update the top-level `README.md` so that other users can easily find the new example along side the other examples?  See https://github.com/confluentinc/kafka-streams-examples#available-examples.  I'd use words like "message routing", "dynamic routing" for the "Concepts used" column, for instance.

ybyzek:
@miguno 👍

PR #199, opened 2019-02-12
Various minor cleanups including using the latest APIs

miguno:
I covered many examples but not all of them.

### DIFF
--- a/src/main/java/io/confluent/examples/streams/microservices/PostOrdersAndPayments.java
+++ b/src/main/java/io/confluent/examples/streams/microservices/PostOrdersAndPayments.java
@@ -21,6 +21,8 @@ import javax.ws.rs.client.ClientBuilder;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.Response;
+
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -28,7 +30,6 @@ import java.util.Properties;
 import java.util.Random;
 
 import static io.confluent.examples.streams.microservices.domain.beans.OrderId.id;
-import static io.confluent.examples.streams.microservices.util.MicroserviceUtils.MIN;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON_TYPE;
 
 public class PostOrdersAndPayments {
@@ -105,7 +106,7 @@ public class PostOrdersAndPayments {
             // GET the bean back explicitly
             System.out.printf("Getting order from: %s   .... ", path.urlGet(i));
             returnedOrder = client.target(path.urlGet(i))
-                                  .queryParam("timeout", MIN / 2)
+                                  .queryParam("timeout", Duration.ofMinutes(1).toMillis() / 2)
                                   .request(APPLICATION_JSON_TYPE)
                                   .get(newBean());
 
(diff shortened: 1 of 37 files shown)


src/main/java/io/confluent/examples/streams/microservices/PostOrdersAndPayments.java
```diff
@@ -105,7 +106,7 @@ public static void main(final String[] args) throws Exception {
             // GET the bean back explicitly
             System.out.printf("Getting order from: %s   .... ", path.urlGet(i));
             returnedOrder = client.target(path.urlGet(i))
-                                  .queryParam("timeout", MIN / 2)
+                                  .queryParam("timeout", Duration.ofMinutes(1).toMillis() / 2)
```
mjsax: Why divide by 2? Just use `Duration.ofSeconds(30)` instead ?

src/test/java/io/confluent/examples/streams/StreamToStreamJoinIntegrationTest.java
```diff
@@ -112,10 +111,19 @@ public void shouldJoinTwoStreams() throws Exception {
...
-        JoinWindows.of(TimeUnit.SECONDS.toMillis(5))
+      incidents,
+      (impressionValue, clickValue) ->
+        (clickValue == null)? impressionValue + "/not-clicked-yet": impressionValue + "/" + clickValue,
+      // KStream-KStream joins are always windowed joins, hence we must provide a join window.
+      JoinWindows.of(Duration.ofSeconds(5)),
+      // In this specific example, we don't need to define join serdes explicitly because the key, left value, and
+      // right value are all of type String, which matches our default serdes configured for the application.  However,
+      // we want to showcase the use of `Joined.with(...)` in case your code needs a different type setup.
+      Joined.with(
```
mjsax: Don't think we need this -- default serdes in configs is `StringSerde` anyway.

src/test/java/io/confluent/examples/streams/microservices/OrdersServiceTest.java
```diff
@@ -81,7 +81,7 @@ public void shouldPostOrderAndGetItBack() {
 
     //When GET the bean back via it's location
     OrderBean returnedBean = client.target(response.getLocation())
-        .queryParam("timeout", MIN / 2)
+        .queryParam("timeout", Duration.ofMinutes(1).toMillis() / 2)
```
mjsax: as above. (more below)

src/test/java/io/confluent/examples/streams/StreamToStreamJoinIntegrationTest.java
miguno: I'd like to keep this included because we often get questions from people because they joins failed due to serdes not having being specified, and it's not obvious when/where this is needed.

src/main/java/io/confluent/examples/streams/microservices/PostOrdersAndPayments.java
miguno: ack

src/test/java/io/confluent/examples/streams/microservices/OrdersServiceTest.java
miguno: ack

miguno:
All comments addressed, please review again.

PR #205, opened 2019-03-05
MINOR: Retry topic creation after a race condition to delete then create

bbejeck:
From looking at the test logs, it looks like there can be a race condition from when the topic deletion is completed and the request to create the topic `orders`
This line errors out because the topic exists
```
[2019-03-05 15:25:40,982] INFO [kafka-request-handler-1] [Admin Manager on Broker 0]: Error processing create topic request for topic orders with arguments (numPartitions=1, replicationFactor=1, replicasAssignments={}, configs={}) (kafka.server.AdminManager)
org.apache.kafka.common.errors.TopicExistsException: Topic 'orders' already exists.
```
However on the very next line we can see the deletion completes successfully
```
[2019-03-05 15:25:41,010] INFO [controller-event-thread] [Topic Deletion Manager 0], Deletion of topic orders successfully completed (kafka.controller.TopicDeletionManager)
```

So I propose if we get an `TopicExistsException` we re-try once to account for this race condition.

### DIFF
--- a/src/test/java/io/confluent/examples/streams/microservices/OrdersServiceTest.java
+++ b/src/test/java/io/confluent/examples/streams/microservices/OrdersServiceTest.java
@@ -9,6 +9,7 @@ import io.confluent.examples.streams.microservices.domain.beans.OrderBean;
 import io.confluent.examples.streams.microservices.util.MicroserviceTestUtils;
 import io.confluent.examples.streams.microservices.util.Paths;
 
+import org.apache.kafka.common.errors.TopicExistsException;
 import org.apache.kafka.test.TestUtils;
 import org.junit.After;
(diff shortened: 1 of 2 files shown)


mjsax:
The overall scenario is unclear to me. Does the test fail if a `TopicExistsException` is thrown? Also, should `deleteTopicsAndWait()` ensure that the topic is deleted properly? If it returns and the topic was not deleted yet, we might want to fix `deleteTopicsAndWait()` instead?

bbejeck:
@mjsax Actually this race condition is a known issue with the `AdminClient` https://issues.apache.org/jira/browse/KAFKA-6098 

I'm closing this PR in favor of opening a new one to use logic present in AK integration tests

PR #206, opened 2019-03-07
MINOR: Need to use ZK client on broker to confirm topics deleted

bbejeck:
We've recently started seeing errors in the examples test due to race condition when using the `AdminClient` to confirm if a topic has been deleted.

The failure shows up in the logs with the failure to create the topic as it still exists as the test condition returned `true`
```
[2019-03-06 01:55:15,914] INFO [data-plane-kafka-request-handler-0] [Admin Manager on Broker 0]: Error processing create topic request for topic orders with arguments (numPartitions=1, replicationFactor=1, replicasAssignments={}, configs={}) (kafka.server.AdminManager)
org.apache.kafka.common.errors.TopicExistsException: Topic 'orders' already exists.
```
And a little later in the log we see the actual deletion taking place
```
[2019-03-06 01:55:15,941] INFO [controller-event-thread] [Topic Deletion Manager 0], Deletion of topic orders successfully completed (kafka.controller.TopicDeletionManager)
```
This PR uses the `KafkaEmbedded` `zk` client to check for topic deletion in the test condition.

### DIFF
--- a/src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
+++ b/src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
@@ -19,24 +19,20 @@ import io.confluent.examples.streams.zookeeper.ZooKeeperEmbedded;
 import io.confluent.kafka.schemaregistry.RestApp;
 import io.confluent.kafka.schemaregistry.avro.AvroCompatibilityLevel;
 import io.confluent.kafka.schemaregistry.rest.SchemaRegistryConfig;
-import kafka.server.KafkaConfig$;
-import org.apache.kafka.clients.admin.AdminClient;
-import org.apache.kafka.clients.admin.AdminClientConfig;
-import org.apache.kafka.clients.admin.ListTopicsResult;
-import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
-import org.apache.kafka.test.TestCondition;
-import org.apache.kafka.test.TestUtils;
-import org.junit.rules.ExternalResource;
-import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
```diff
@@ -254,16 +250,8 @@ private TopicsDeletedCondition(final String... topics) {
 
     @Override
     public boolean conditionMet() {
-      final Properties properties = new Properties();
```
mjsax: Should we keep this old code, put it into comments, and add a comment that we should change this back after this bug is fixed?

bbejeck: @mjsax I'm not a big fan of leaving commented out code even with an explanation, so I've added a `TODO` saying to use `AdminClient` once the issue is fixed.  Does that work for you?

This is just my personal preference so I'll add the code back in comments if you insist.

mjsax: SGTM -- just referencing the ticket is better.

(Just want to make sure, we don't say "hey, why don't we use `AdminClient` here" and reintroduce this issue.)

bbejeck:
merged with pint

src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
vvcephei: :+1: This todo is perfect. When we decide to resolve it, we can check `git blame` to find the prior code if we like.

PR #217, opened 2019-03-26
MINOR: add gc logging to tests

vvcephei:


vvcephei:
Hey @mjsax @bbejeck @ableegoldman ,

Please take a look at this small PR to enable GC logs during tests. I think that this will shed some light on why the tests sometimes time out.

Thanks,
-John

vvcephei:
There was fortunately another flaky test failure with GC enabled in this PR build, and I determined that GC was not the cause.

Actually, it looks like the connection failures I was observing in the tests are a result of the broker shutting down before Streams finishes shutdown, *after* the test has already failed.

I'm going to make some further changes to help further debugging efforts, but it'll take a while. Please hold...

guozhangwang:
@vvcephei is it ready for reviews now?

vvcephei:
I'm going to trash this PR in favor of switching to TopologyTestDriver in post-2.0 branches.

PR #219, opened 2019-04-02
Example to test actual implementation with TopologyTestDriver

jukkakarvanen:
Readme states: "These examples are also a good starting point to learn how to implement your own end-to-end integration tests.". I disagree. See testing instructions:
https://docs.confluent.io/current/streams/developer-guide/test-streams.html

Now the Integration Tests are mainly copy/paste of actual implementation and not testing actual implementation class at all. The most of the test could also be done without EmbeddedKafka with TopologyTestDriver. 

In this pull request there is a new WordCountLambdaExampleTest added utilizing TopologyTestDriver.
To be able to test WordCountLambdaExample implementation of WordCountLambdaExample main method needed to be refactored, so actual configuration and Stream building are own methods.

WordCountLambdaIntegrationTest class is also modified to utilize Stream logic of actual application and that way avoid code duplication.

The same logic should be applied also to other tests.


jukkakarvanen:
[clabot:check]

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
```diff
@@ -0,0 +1,137 @@
...
+public class WordCountLambdaExampleTest {
+
+  private TopologyTestDriver testDriver;
+
+  private StringDeserializer stringDeserializer = new StringDeserializer();
+  private LongDeserializer longDeserializer = new LongDeserializer();
+  private ConsumerRecordFactory<String, String> recordFactory = new ConsumerRecordFactory<>(new StringSerializer(), new StringSerializer());
+
+  @Before
+  public void setup() throws IllegalAccessException, ClassNotFoundException, InstantiationException {
```
bbejeck: Can remove the `throws` clause and all exceptions

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Stream processing unit test of {@link WordCountLambdaExample}, using TopologyTestDriver.
+ *
+ * @author Jukka Karvanen / jukinimi.com
```
bbejeck: nit: we usually don't have any author names in the javadoc

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
```diff
@@ -0,0 +1,137 @@
...
+      testDriver.close();
+    } catch (final RuntimeException e) {
+      // https://issues.apache.org/jira/browse/KAFKA-6647 causes exception when executed in Windows, ignoring it
+      // Logged stacktrace cannot be avoided
+      System.out.println("Ignoring exception, test failing in Windows due this exception:" + e.getLocalizedMessage());
+    }
+  }
+
+
+  /** Read one Record from output topic.
```
bbejeck: nit: move the text one line below the start of the javadoc

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
```diff
@@ -0,0 +1,137 @@
...
+   *
+   * @return ProducerRecord containing word as key and count as value
+   */
+  private ProducerRecord<String, Long> readOutput() {
+    return testDriver.readOutput(WordCountLambdaExample.outputTopic, stringDeserializer, longDeserializer);
+  }
+
+  /** Read counts from output to map.
```
bbejeck: ditto here and elsewhere below

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.confluent.examples.streams;
+
+import org.apache.kafka.clients.producer.ProducerRecord;
```
bbejeck: nit: remove wildcard imports

src/main/java/io/confluent/examples/streams/WordCountLambdaExample.java
```diff
@@ -111,8 +111,48 @@
  */
 public class WordCountLambdaExample {
 
+  static final String inputTopic = "inputTopic";
+  static final String outputTopic = "outputTopic";
+
   public static void main(final String[] args) throws Exception {
```
bbejeck: nit: can rid of `throws Exception`

src/main/java/io/confluent/examples/streams/WordCountLambdaExample.java
```diff
@@ -111,8 +111,48 @@
  */
 public class WordCountLambdaExample {
 
+  static final String inputTopic = "inputTopic";
```
bbejeck: These topic names need to match the names of the directions in the javadoc
`inputTopic`= `streams-plaintext-input`
`outputTopic` = `streams-wordcount-output`

README.md
```diff
@@ -43,11 +43,12 @@ There are two kinds of examples:
   must first install and run Apache Kafka and friends, which we describe in section
   [Packaging and running the examples](#packaging-and-running).  Each example also states its exact requirements and
   instructions at the very top.
-* **Examples under [src/test/](src/test/)**: These examples are a bit longer because they implement integration tests
-  that demonstrate end-to-end data pipelines.  Here, we use a testing framework to automatically spawn embedded Kafka
+* **Examples under [src/test/](src/test/)**: These examples should test applications under [src/main/](src/main/).
+  Unit Tests with TopologyTestDriver test the stream logic without external system dependencies.
+  The integration tests use a testing framework to automatically spawn embedded Kafka
```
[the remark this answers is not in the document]
jukkakarvanen: Changed.

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
jukkakarvanen: Done

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
jukkakarvanen: Removed

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
jukkakarvanen: Done

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
jukkakarvanen: Done

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
jukkakarvanen: Ok, Changed IDEA settings.

src/main/java/io/confluent/examples/streams/WordCountLambdaExample.java
jukkakarvanen: Removed (Old problem,)

src/main/java/io/confluent/examples/streams/WordCountLambdaExample.java
jukkakarvanen: Good catch. Fixed.

src/test/java/io/confluent/examples/streams/WordCountLambdaExampleTest.java
jukkakarvanen: Done

src/main/java/io/confluent/examples/streams/WordCountLambdaExample.java
jukkakarvanen: Problem was inherited from WordCountLambdaIntegrationTest.java

jukkakarvanen:
No idea why previous test failed due to WordCountScalaIntegrationTest failure and now it is successful again.

vvcephei:
These integration tests are quite flaky, which is part of my motivation to just switch to TopologyTestDriver.

jukkakarvanen:
> You have an excellent point. I'm wondering if we should even keep the integration test at all. The application logic of the WordCountExample depends purely on Streams; it has no logical interactions with the broker. Streams already guarantees that it works with the broker, so it seems like the only necessary test is to verify the application logic, which is better done with the TopologyTestDriver.
> 
> What do you think about removing the integration test entirely?
> 
@vvcephei, Yes, I don't see the point of the this kind of integration test with EmbeddedKafka at all, if you want to test Kafka Stream application logic. I understand the use case of EmbeddedSingleNodeKafkaCluster if the application contains also consumers or producers, but not in pure Kafka Streams.

Maybe somebody knowing the internals of the TopologyTestDriver could comment is there possibility of difference, but I have found TopologyTestDriver a fast and reliable way of the testing the Kafka Stream application.

mjsax:
It's historical reasons. When the examples where created, there was no `TopologyTestDriver` yet. And we never invested time to rewrite the tests later.

jukkakarvanen:
FYI:  There are PR https://github.com/apache/kafka/pull/6569 to some point of time get rid of try / catch in tearDown, I needed to add because I am working with Windows.

I also working to propose some extra Helper class to make simplify the use of TopologyTestDriver to be added kafka-streams-test-utils (and possible independent package for old Kafka versions)

I added an example of how those class could be used in this WordCountLambdaExampleTest in separate brach of my repo: https://github.com/jukkakarvanen/kafka-streams-examples/compare/TopologyTestDriver_tests...jukkakarvanen:InputOutputTopic

vvcephei:
Thanks @jukkakarvanen ,

It sounds like we can actually go ahead and just get rid of the `WordCountScalaIntegrationTest` and just use your new test instead.

Thanks also for those other PRs. I'd say, we should just consider them all independently. If we merge the other ones first, we can update this PR, but if we merge this PR first, we can always use a new PR to improve the test further.

bbejeck:
@jukkakarvanen thanks for the contribution!
To follow up on @vvcephei's comment If we can get rid of the `WordCountScalaIntegrationTest` in this PR then we can just the new test instead

jukkakarvanen:
@bbejeck  `WordCountScalaIntegrationTest` class and references to it removed

bbejeck:
I pulled down the PR and verified the `WordCountLambdaExample`

bbejeck:
Thanks @jukkakarvanen !

bbejeck:
merged with pint

mjsax:
Sorry that I am late to the game. `TopologyTestDriver` was introduced in AK 1.1, thus I am wondering if this should have gone to `4.1.0-post`? Maybe we can still backport it?

PR #235, opened 2019-04-17
Add integration test examples for reduce() and aggregate() in Java and Scala

miguno:
Edit:  Here's an updated summary of this PR.  I apologize that it is not "one-PR-per-addressed-issue", but there was so much overlap that I didn't want to waste sweat and tears to split it into many smaller PRs.

- Add reduce() and aggregate() examples in Java and Scala, using our new TopologyTestDriver setup.
- Add a "how to rekey a table" example.
- Update existing Scala examples to use TopologyTestDriver, e.g. `ProbabilisticCountingScalaIntegrationTest`.
- Introduced a new class `IntegrationTestScalaUtils` so that writing test input and reading test output doesn't require type-conversion shenanigans for primitive types anymore (e.g. to force a Scala `Int` to be treated as a Java `Integer`), using implicits. This makes the Scala tests much smoother to read, similar to how the Scala DSL doesn't force you to specify explicit serdes for primitives.
- Update README to distinguish between TopologyTestDriver examples and "embedded cluster" examples.
- Minor refactoring of existing examples to simplify the code, e.g. by switching from `produceKeyValuesSynchronously()` to `produceValuesSynchronously()` when the input data is values-only. Also, I consolidated some examples (where I had the time) so their code is more consistent across each other.
- Fixed a bug in `CustomStreamTableJoinIntegrationTest` along the way where the test output wasn't read and validated as it should.


miguno:
This will allow us to point users that have questions around aggregation (reduce, aggregate) to ready-to-use examples that work end to end.

src/test/java/io/confluent/examples/streams/AggregateIntegrationTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * End-to-end integration test that shows how to aggregate messages via `groupBy()` and `aggregate()`.
+ *
+ * Note: This example uses lambda expressions and thus works with Java 8+ only.
+ */
+public class AggregateIntegrationTest {
+
+  @ClassRule
+  public static final EmbeddedSingleNodeKafkaCluster CLUSTER = new EmbeddedSingleNodeKafkaCluster();
```
mjsax: We started to rewrite out test to use `TopologyTestDriver` whenever possible. Can we update this PR accordingly?

miguno:
@mjsax : I refactored one of the four examples, `ReduceTest` (formerly: `ReduceIntegrationTest`) to use `TopologyTestDriver`. See commit https://github.com/confluentinc/kafka-streams-examples/pull/235/commits/3316096e85b8d8729b6a16d0aae6ab529887fa6c.

Please review the refactored test.  If you are +1, I will similarly refactor the other three examples in this PR.

miguno:
retest this please

miguno:
Note: This PR is currently on-hold until the PR https://github.com/confluentinc/kafka-streams-examples/pull/248 by @vvcephei is merged, because this PR should follow the same style and re-use any new code introduced by https://github.com/confluentinc/kafka-streams-examples/pull/248.

vvcephei:
@miguno , my other PR is merged now, and I'm porting to the 5.2.0-post branch right now. You should be unblocked.

miguno:
Thanks, @vvcephei. This PR is now ready to review.

Here's an updated summary of what is now covered in this PR.  I apologize that it is not "one-PR-per-addressed-issue", but there was so much overlap that I didn't want to waste sweat and tears to split it into many smaller PRs.

- Add reduce() and aggregate() examples in Java and Scala, using our new TopologyTestDriver setup.
- Update existing Scala examples to use TopologyTestDriver, e.g. `ProbabilisticCountingScalaIntegrationTest`.
- Introduced a new class `IntegrationTestScalaUtils` so that writing test input and reading test output doesn't require type-conversion shenanigans for primitive types anymore (e.g. to force a Scala `Int` to be treated as a Java `Integer`), using implicits. This makes the Scala tests much smoother to read, similar to how the Scala DSL doesn't force you to specify explicit serdes for primitives.
- Update README to distinguish between TopologyTestDriver examples and "embedded cluster" examples.
- Minor refactoring of existing examples to simplify the code, e.g. by switching from `produceKeyValuesSynchronously()` to `produceValuesSynchronously()` when the input data is values-only. Also, I consolidated some examples (where I had the time) so their code is more consistent across each other.
- Fixed a bug in `CustomStreamTableJoinIntegrationTest` along the way where the test output wasn't read and validated as it should.

src/test/java/io/confluent/examples/streams/AggregateTest.java
```diff
@@ -0,0 +1,116 @@
...
+        );
+    aggregated.toStream().to(outputTopic, Produced.with(Serdes.String(), Serdes.Long()));
+    return builder;
+  }
+
+  private Properties createTopologyConfiguration() {
+    final Properties streamsConfiguration = new Properties();
+    streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "aggregate-test");
+    streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
+    streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
vvcephei: this config is unused in TTD

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -397,21 +398,43 @@ private static boolean continueConsuming(final int messagesConsumed, final int m
...
+   * Like {@link IntegrationTestUtils#produceValuesSynchronously(String, Collection, Properties)}, except for use with
+   * TopologyTestDriver tests, rather than "native" Kafka broker tests.
+   *
+   * @param topic              Kafka topic to write the data records to
+   * @param values             Message values to write to Kafka (keys will have type byte[] and be set to null)
+   * @param topologyTestDriver The {@link TopologyTestDriver} to send the data records to
+   * @param valueSerializer    The {@link Serializer} corresponding to the value type
+   * @param <V>                Value type of the data records
+   */
+  static <K, V> void produceValuesSynchronously(final String topic,
```
vvcephei: This is a nice addition, but can you target this change to `4.1.0-post`? Otherwise, every change to every test is going to run into merge conflicts when they hit this branch.

vvcephei: Note, offhand, it seems like this whole PR could be targeted to `4.1.0-post`

src/test/java/io/confluent/examples/streams/MapFunctionLambdaIntegrationTest.java
```diff
@@ -33,7 +33,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
- * End-to-end integration test based on MapFunctionLambdaExample, using an embedded Kafka cluster.
+ * End-to-end integration test based on MapFunctionLambdaExample, using TopologyTestDriver.
```
vvcephei: I guess it's not an end-to-end test at all (and never was). Maybe the Javadoc doesn't have to state things that are readily apparent from the code anyway.

src/test/java/io/confluent/examples/streams/ReduceTest.java
```diff
@@ -0,0 +1,112 @@
...
+        new KeyValue<>(456, "all"),
+        new KeyValue<>(123, "kafka"),
+        new KeyValue<>(456, "the"),
+        new KeyValue<>(456, "things"),
+        new KeyValue<>(123, "streams")
+    );
+
+    final Map<Integer, String> expectedOutput = new HashMap<>();
+    expectedOutput.put(456, "stream all the things");
+    expectedOutput.put(123, "hello world kafka streams");
```
vvcephei: nifty!

src/test/java/io/confluent/examples/streams/ReduceTest.java
```diff
@@ -0,0 +1,112 @@
...
+      assertThat(actualOutput).hasSameSizeAs(expectedOutput);
+      assertThat(actualOutput).containsAllEntriesOf(expectedOutput);
+    }
+  }
+
+  private StreamsBuilder createTopology() {
+    final StreamsBuilder builder = new StreamsBuilder();
+    final KStream<Integer, String> input = builder.stream(inputTopic, Consumed.with(Serdes.Integer(), Serdes.String()));
+    final KTable<Integer, String> concatenated =
+        input.groupByKey(Grouped.with(Serdes.Integer(), Serdes.String())).reduce((v1, v2) -> v1 + " " + v2);
```
vvcephei: Similar to the other example, do you mind giving a textual description of the app? It's just a little distracting to have to figure out what it's doing, when the point is really how to test it.

src/test/java/io/confluent/examples/streams/ReduceTest.java
```diff
@@ -0,0 +1,112 @@
...
+    concatenated.toStream().to(outputTopic, Produced.with(Serdes.Integer(), Serdes.String()));
+    return builder;
+  }
+
+  private Properties createTopologyConfiguration() {
+    final Properties streamsConfiguration = new Properties();
+    streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "reduce-test");
+    streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
+    streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
vvcephei: Also here, this config shouldn't be necessary.

src/test/java/io/confluent/examples/streams/TableRekeyingTest.java
```diff
@@ -0,0 +1,138 @@
...
+    // Variant 1 (https://docs.confluent.io/current/streams/faq.html#option-1-write-kstream-to-ak-read-back-as-ktable)
+    // Here, we re-key the KTable, write the results to a new topic, and then re-read that topic into a new KTable.
+    table
+        .toStream()
+        .map((key, value) -> KeyValue.pair(value, key))
+        .to(outputTopic1, Produced.with(Serdes.String(), Serdes.Integer()));
+    final KTable<String, Integer> rekeyedTable1 =
+        builder.table(outputTopic1, Consumed.with(Serdes.String(), Serdes.Integer()));
+
+    // Variant 2 (https://docs.confluent.io/current/streams/faq.html#option-2-perform-a-dummy-aggregation)
```
vvcephei: This one doesn't work, actually, because reduce bypasses incoming null values (tombstones), instead of passing them into the function, so the resulting table will have "zombie" records that should have been deleted.

vvcephei: should be possible to exercise this by sending an input record and then sending a tombstone for the same key.

src/test/java/io/confluent/examples/streams/WordCountLambdaIntegrationTest.java
```diff
@@ -59,78 +56,69 @@
...
-      assertThat(actualWordCounts).isEqualTo(expectedWordCounts);
-    }
+    final Properties streamsConfiguration = new Properties();
+    streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "wordcount-lambda-integration-test");
+    streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
+    streamsConfiguration.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
+    streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
+    // The commit interval for flushing records to state stores and downstream must be lower than
+    // this integration test's timeout (30 secs) to ensure we observe the expected processing results.
+    streamsConfiguration.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, TimeUnit.SECONDS.toMillis(10));
```
vvcephei: this config should not be necessary with TTD.

src/test/scala/io/confluent/examples/streams/IntegrationTestScalaUtils.scala
```diff
@@ -0,0 +1,104 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.confluent.examples.streams
+
+import org.apache.kafka.common.serialization.{Deserializer, Serde, Serializer, Serdes => JSerdes}
+import org.apache.kafka.streams.kstream.{Windowed, WindowedSerdes}
+import org.apache.kafka.streams.{KeyValue, TopologyTestDriver}
+
+object IntegrationTestScalaUtils {
```
vvcephei: Might be worth some comments explaining what's in here, and why it's needed.

src/test/java/io/confluent/examples/streams/AggregateTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    try (final TopologyTestDriver testDriver = new TopologyTestDriver(builder.build(), streamsConfiguration)) {
+      // Step 2: Write the input
+      IntegrationTestUtils.produceValuesSynchronously(inputTopic, inputValues, testDriver, new StringSerializer());
+
+      // Step 3: Validate the output
+      final Map<String, Long> actualOutput = IntegrationTestUtils.drainTableOutput(
+          outputTopic, testDriver, new StringDeserializer(), new LongDeserializer());
+      assertThat(actualOutput).hasSameSizeAs(expectedOutput);
+      assertThat(actualOutput).containsAllEntriesOf(expectedOutput);
```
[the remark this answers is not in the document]
miguno: Ack. From another test scenario (not this PR) I thought to remember that `assertThat().isEqualsTo()` would not cover all possible ways the two maps could be different (e.g. one map having the same entries as the other but also an additional entry). Guess I was wrong :)

src/test/java/io/confluent/examples/streams/AggregateTest.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  private StreamsBuilder createTopology() {
+    final StreamsBuilder builder = new StreamsBuilder();
+    final KStream<byte[], String> input = builder.stream(inputTopic, Consumed.with(Serdes.ByteArray(), Serdes.String()));
```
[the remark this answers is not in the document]
miguno: Hmm, what I don't like about `NothingSerde<T>` is that it is not a default KStreams serde.  I think it is ok to use it when writing test data (via `IntegrationTestUtils`) and reading test output (via `IntegrationTestUtils`), but I am not convinced we should use it for the "real" example code. (For instance, you cannot easily copy-paste such snippets into your own code, or into a presentation slide, because outside of `kafka-streams-examples` the code does not work.)

src/test/java/io/confluent/examples/streams/AggregateTest.java
```diff
@@ -0,0 +1,116 @@
...
+    final KTable<String, Long> aggregated = input
+        .groupBy(
+            (key, value) -> (value != null && value.length() > 0) ? value.substring(0, 1).toLowerCase() : "",
```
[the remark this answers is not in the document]
miguno: Ack.

src/test/java/io/confluent/examples/streams/AggregateTest.java
miguno: Ack

src/test/java/io/confluent/examples/streams/MapFunctionLambdaIntegrationTest.java
miguno: Renamed to "Test based on MapFunctionLambdaExample, using TopologyTestDriver."

src/test/java/io/confluent/examples/streams/ReduceTest.java
miguno: Ack

src/test/scala/io/confluent/examples/streams/IntegrationTestScalaUtils.scala
miguno: Ack

src/test/java/io/confluent/examples/streams/WordCountLambdaIntegrationTest.java
miguno: Ack

src/test/java/io/confluent/examples/streams/ReduceTest.java
miguno: Ack

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
miguno: > This is a nice addition, but can you target this change to 4.1.0-post? Otherwise, every change to every test is going to run into merge conflicts when they hit this branch.

Why would that be so?  The changes in this PR are targeting `5.2.0`, and would therefore not be pushed into any `4.x` branches?

Also, I tried rebasing this PR on `4.1.0-post` for the fun of it, but there are 120+ commits, many of which result into merge failures.  I don't think we need to retroactively also change the `4.x` ones.

src/test/java/io/confluent/examples/streams/TableRekeyingTest.java
miguno: Hmm, there's actually more wrong this example -- even Variant 1 doesn't work. I'll remove this example from the PR and will follow-up later.

miguno:
@vvcephei: Thanks for the review. I addressed all your comments with the following exceptions:

- For reasons stated in https://github.com/confluentinc/kafka-streams-examples/pull/235#discussion_r286795231, I don't want to use `NothingSerde` in the specific situation that you commented on.
- I don't want to re-target this PR to `4.1.0-post` because doing so will be a lot of work (160+ commits of which many result in merge conflicts), see discussion at https://github.com/confluentinc/kafka-streams-examples/pull/235#discussion_r286641308.
- I removed `TableRekeyingTest` from the PR.  I'll need to think about how to actually do this, because neither of the variants actually work with tombstones.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: That makes sense. In that case, I would recommend reverting the "cleanup" portions of this PR... The main problem is that every time I have to make a change to the examples in an older branch, I already spend hours resolving merge conflicts. If we're going to make cross-cutting changes, they really need to be done in the first applicable branch.

On the other hand, your main goal here is to add some new examples, which is not cross-cutting, and I'd have no problem just targeting 5.2.0 for that.

src/test/java/io/confluent/examples/streams/AggregateTest.java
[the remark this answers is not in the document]
vvcephei: This is a very good point... I'll make a ticket to add the NothingSerde to Serdes.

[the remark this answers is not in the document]
vvcephei: shaboom: https://issues.apache.org/jira/browse/KAFKA-8455

Thanks for the feedback, I agree 100%

src/test/java/io/confluent/examples/streams/TableRekeyingTest.java
vvcephei: Yeah, upon reflection, this isn't really just "rekeying" a table. It's more like a 1:many query, which would need an approach similar to https://github.com/apache/kafka/pull/5527 to be done correctly.

To extract the problem, say you have a table:
```
Table1:
1 -> A
2 -> A
```

And you want to invert the k/v relationship. Table1 isn't a bijection, so when we interpret "A" as a key, it actually has 2 values, not one, so the proper representation of the rekeyed table is:

```
Table2:
A -> {1,2}
```

Then, when we run the following updates on Table1, we should still get the right result for Table2:
```
put(2,B);

Table1:
1 -> A
2 -> B

implicit:( subtract(A, 2), add(B,2) )

Table2:
A -> {1}
B -> {2}
```
```
put(3,A);

Table1:
1 -> A
2 -> B
3 -> A

implicit:( add(A,3) )

Table2:
A -> {1,3}
B -> {2}
```

Does this seem right?

vvcephei: Oh, yeah, I forgot about tombstones, but that should also work. Picking up from the last state:
```
delete(1);

Table1:
2 -> B
3 -> A

implicit:( subtract(A,1) )

Table2:
A -> {3}
B -> {2}
```

vvcephei: The real brain twister is if you think of inverting Table2. Theoretically, you should get Table1 again, but there's no enforced uniqueness guarantee. Still, you have a priori knowledge that Table1:Table2 is 1:many, and therefore Table2:Table1 is many:1.

The question is whether you could naively implement the aggregation just to take every update as the aggregation result, with the following aggregation algorithm:
```
initializer: () -> null
subtractor: (key, value, agg) -> null
adder: (key, value, agg) -> value
```

Not sure if this results in a sound re-inversion. It might be interesting to test it out.

miguno:
@vvcephei :

> In that case, I would recommend reverting the "cleanup" portions of this PR... The main problem is that every time I have to make a change to the examples in an older branch, I already spend hours resolving merge conflicts. If we're going to make cross-cutting changes, they really need to be done in the first applicable branch.

Which changes in this PR would you like me to remove?

vvcephei:
Hey @miguno , so sorry for the delay. What I meant was essentially to strip out anything that's not directly related to your new examples.

e.g., if you just add the new examples, and your new IntegrationTestUtils convenience function, but don't migrate anything else, it will merge cleanly to future changes in older branches.

miguno:
https://github.com/confluentinc/kafka-streams-examples/pull/235#issuecomment-502128782
> e.g., if you just add the new examples, and your new IntegrationTestUtils convenience function, but don't migrate anything else, it will merge cleanly to future changes in older branches.

This is now available as a smaller-scoped PR at https://github.com/confluentinc/kafka-streams-examples/pull/270

miguno:
> - Fixed a bug in `CustomStreamTableJoinIntegrationTest` along the way where the test output wasn't read and validated as it should.

This has since been covered in PR https://github.com/confluentinc/kafka-streams-examples/pull/296.

miguno:
> - Update existing Scala examples to use TopologyTestDriver, e.g. `ProbabilisticCountingScalaIntegrationTest`.

`ProbabilisticCountingScalaIntegrationTest` is still not using TTD, but it was since updated to use our Scala test utils, see PR https://github.com/confluentinc/kafka-streams-examples/pull/275.

miguno:
All the major items of this PR have since been covered by other, smaller PRs.  Closing this.

PR #239, opened 2019-04-21
MINOR: fix indention

mjsax:


src/main/java/io/confluent/examples/streams/microservices/domain/Schemas.java
```diff
@@ -1,20 +1,21 @@
...
+import org.apache.kafka.common.serialization.Serdes;
+
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import org.apache.kafka.common.serialization.Serde;
-import org.apache.kafka.common.serialization.Serdes;
+
+import static io.confluent.examples.streams.microservices.util.MicroserviceUtils.ProductTypeSerde;
+import static io.confluent.kafka.serializers.AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG;
```
mjsax: @cadonna Just realizing that this should not happen -- using wrong styleguide triggered auto-reordering. Reverting this part.

PR #242, opened 2019-04-22
MINOR: Use TopologyTestDriver for UserCountsPerRegionLambdaIntegrationTest

vvcephei:
Migrate `UserCountsPerRegionLambdaIntegrationTest` to TopologyTestDriver. The first of many, others will follow.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,6 +87,7 @@
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-streams-test-utils</artifactId>
             <version>${kafka.version}</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.avro</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -84,6 +84,11 @@
             <version>${kafka.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams-test-utils</artifactId>
+            <version>${kafka.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.avro</groupId>
             <artifactId>avro</artifactId>
             <version>${avro.version}</version>
--- a/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
+++ b/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
@@ -22,27 +22,29 @@ import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.record.TimestampType;
+import org.apache.kafka.common.serialization.Deserializer;
+import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.streams.KafkaStreams;
 import org.apache.kafka.streams.KeyValue;
+import org.apache.kafka.streams.TopologyTestDriver;
 import org.apache.kafka.streams.errors.InvalidStateStoreException;
 import org.apache.kafka.streams.state.QueryableStoreType;
 import org.apache.kafka.streams.state.ReadOnlyKeyValueStore;
 import org.apache.kafka.streams.state.ReadOnlyWindowStore;
 import org.apache.kafka.streams.state.WindowStoreIterator;
 import org.apache.kafka.test.TestUtils;
-import org.junit.Test;
 
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.stream.Collectors;
-
-import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * Utility functions to make integration testing more convenient.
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -83,6 +83,11 @@
             <artifactId>kafka-streams</artifactId>
             <version>${kafka.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-streams-test-utils</artifactId>
+            <version>${kafka.version}</version>
+        </dependency>
```
mjsax: Should this be in testing scope only?

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -277,4 +279,87 @@ private static boolean continueConsuming(final int messagesConsumed, final int m
...
+      if (record == null) {
+        break;
+      } else {
+        results.put(record.key(), record.value());
+      }
+    }
+    return results;
+  }
+
+  static <K, V> void produceKeyValuesSynchronously(final String inputTopic,
```
mjsax: Do we need to add `Synchronously` to the name? `TopologyTestDriver` does not support any async stuff anyway.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -277,4 +279,87 @@ private static boolean continueConsuming(final int messagesConsumed, final int m
...
+        0,
+        0,
+        0,
+        TimestampType.CREATE_TIME,
+        -1,
+        -1,
+        -1,
+        keySerializer.serialize(inputTopic, entity.key),
+        valueSerializer.serialize(inputTopic, entity.value)
+      ));
```
mjsax: Use constants, like `NO_TIMESTAMP`, `NO_CHECKSUM` and similar. For others (with no available consts, add comment?) -- Also, there are multiple constructors -- can't we use one with less parameters?

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -277,4 +279,87 @@ private static boolean continueConsuming(final int messagesConsumed, final int m
         30000,
         "Expected values not found in WindowStore"); }
 
+  static <K, V> Map<K, V> drainTableOutput(final String outputTopic,
+                                           final TopologyTestDriver topologyTestDriver,
+                                           final Deserializer<K> keyDeserializer,
+                                           final Deserializer<V> valueDeserializer) {
+
+    final Map<K, V> results = new LinkedHashMap<>();
```
mjsax: Why `LinkedHashMap` ?

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -277,4 +279,87 @@ private static boolean continueConsuming(final int messagesConsumed, final int m
...
+   * Creates a map from a sequence of entries
+   *
+   * @param entries The entries to map
+   * @param <K>     The key type
+   * @param <V>     The value type
+   * @return A map
+   */
+  @SafeVarargs
+  static <K, V> Map<K, V> mkMap(final Map.Entry<K, V>... entries) {
+    final LinkedHashMap<K, V> result = new LinkedHashMap<>();
```
mjsax: Why `LinkedHashMap`?

src/test/java/io/confluent/examples/streams/UserCountsPerRegionLambdaIntegrationTest.java
```diff
@@ -15,83 +15,73 @@
...
 
+import static io.confluent.examples.streams.IntegrationTestUtils.mkEntry;
+import static io.confluent.examples.streams.IntegrationTestUtils.mkMap;
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * End-to-end integration test that demonstrates how aggregations on a KTable produce the expected
  * results even though new data is continuously arriving in the KTable's input topic in Kafka.
- *
+ * <p>
```
mjsax: Thanks for the cleanup!

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -277,4 +279,87 @@ private static boolean continueConsuming(final int messagesConsumed, final int m
         30000,
         "Expected values not found in WindowStore"); }
 
+  static <K, V> Map<K, V> drainTableOutput(final String outputTopic,
```
mjsax: Can we add a JavaDoc comment here? This method is super useful but maybe not clear what it does if one does not look into the implementation.

pom.xml
vvcephei: yes!

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: Not sure; I left the name so it would clearly be a "peer" to the other `produceKeyValuesSynchronously`. I.e., functionally equivalent, but for TopologyTestDriver instead.

I figured this would help code readability also, so it would continue to be self-documenting for people (esp those not familiar with the TopologyTestDriver) reading the tests.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: To improve readability. This way, when you print out the results (e.g., while writing a test, or playing around with the code, or looking at test failure output), the string form of the results is:
* stable order
* order depends on history of the output topic, so you can reason about what you expect to see

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
mjsax: Ack. Fine with me. Was just a thought.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: Roger using constants.

Sadly, there's no constructor that lets you set topic, partition, and timestamp except for the full constructor.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: Same basic reason: The API is Map, so I'm not guaranteeing anything, but I figured it would help test/debugging ergonomics if the order were stable and depends on the code to create the map.

src/test/java/io/confluent/examples/streams/UserCountsPerRegionLambdaIntegrationTest.java
vvcephei: B) I think this credit goes to IDEA, actually. I'm just an agent to convert IDEA autoformatting to Github pull requests!

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: You bet. Good idea.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
mjsax: Ack. Should it be `Map` left hand side?

vvcephei: Sure!

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: I added docs for the corresponding `produce` method as well. Thanks for pointing this out.

vvcephei:
Thanks for the review, @mjsax !

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -296,24 +315,37 @@ private static boolean continueConsuming(final int messagesConsumed, final int m
...
                                                    final Serializer<K> keySerializer,
                                                    final Serializer<V> valueSerializer) {
-    for (final KeyValue<K, V> entity : entities) {
-      topologyTestDriver.pipeInput(new ConsumerRecord<>(
-        inputTopic,
+    for (final KeyValue<K, V> entity : records) {
+      final ConsumerRecord<byte[], byte[]> consumerRecord = new ConsumerRecord<>(
+        topic,
         0,
         0,
```
mjsax: Should we really set offset zero for all records? Maybe add a static Map from topic -> offset for tracking? (Might require a `reset()` method.)

Or let user pass in start offsets and increment?

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -296,24 +315,37 @@ private static boolean continueConsuming(final int messagesConsumed, final int m
...
         0,
```
mjsax: If we set timestamp to zero for all, this might be bad for testing windows. Should we pass in a timestamp per key-value pair, too? Or similar? Or use `RecordFactory` ?

vvcephei: I figured I'd cross that bridge when I came to a test that needs it. Is it ok with you to leave this as-is for now, and then implement whatever timestamp support is appropriate when we come to a test that needs to verify windows?

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: Not sure. I agree it creates some semantic ambiguity, although it apparently doesn't affect the logic under test. I guess Streams doesn't introspect on the offset, it just expects to get all the records in offset order. If we implement async processing, this may well change.

If it doesn't matter for this test, I think maybe we should just not do anything fancy. If it starts to matter at some point, I think we should revisit this util method then.

mjsax: Fair enough.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
mjsax: Fine with me.

vvcephei:
Thanks @mjsax and @bbejeck ! I'll merge this, and start on Part 2 of 5,000 .

vvcephei:
ugh, I should have sent this to 4.1.0-post. I'll fix it during the downstream merge.

PR #243, opened 2019-04-23
[WIP do not merge] Update examples to use Serde auto-config

mjsax:
This PR is show the changes required to leverage Serde auto-config.

It also exposes couple of issue with the new feature and this PR won't compile until we sync the upstream changes from the AK repo (including follow up bug fixes in AK).

Opened the PR to give an impression on the impact. More details in inline comments.

### DIFF
--- a/src/main/java/io/confluent/examples/streams/GlobalKTablesExample.java
+++ b/src/main/java/io/confluent/examples/streams/GlobalKTablesExample.java
@@ -34,8 +34,6 @@ import org.apache.kafka.streams.kstream.Materialized;
 import org.apache.kafka.streams.kstream.Produced;
 import org.apache.kafka.streams.state.KeyValueStore;
 
-import java.util.Collections;
-import java.util.Map;
 import java.util.Properties;
 
 /**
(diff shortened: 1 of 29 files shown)


src/main/java/io/confluent/examples/streams/GlobalKTablesExample.java
```diff
@@ -137,22 +135,14 @@ public static KafkaStreams createStreams(final String bootstrapServers,
     // Set to earliest so we don't miss any data that arrived in the topics before the process
     // started
     streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+    // Where to find the Confluent schema registry instance(s). Will automatically be forwarded to all Serdes
+    streamsConfiguration.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, schemaRegistryUrl);
```
mjsax: This change is expected.

src/main/java/io/confluent/examples/streams/microservices/FraudService.java
```diff
@@ -100,19 +105,21 @@ private KafkaStreams processStreams(final String bootstrapServers, final String
     //as caching in Kafka Streams will conflate subsequent updates for the same key. Disabling caching ensures
     //we get a complete "changelog" from the aggregate(...) step above (i.e. every input event will have a
     //corresponding output event.
-    final Properties props = baseStreamsConfig(bootstrapServers, stateDir, SERVICE_APP_ID);
-    props.setProperty(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG, "0");
+    final Properties config = baseStreamsConfig(bootstrapServers, stateDir, SERVICE_APP_ID);
+    config.setProperty(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG, "0");
+    config.put(SCHEMA_REGISTRY_URL_CONFIG, schemaRegistryUrl);
```
mjsax: This change is not expected. For this example, all Serdes are manually configured, and we keep this code because the `Serdes` are also used for regular producers/consumers to write input data and read results.

However, even if all Serdes are manually configured, we still need to set the SR config here to not break the code -- without it, the second call to `configure()` remove the SR config and we fail with an exception at runtime, because the Serdes are not configures correctly and cannot connect to the SR. In fact, it fails with a NPE (cf. https://github.com/confluentinc/schema-registry/issues/1085)

mjsax:
Without https://github.com/apache/kafka/pull/6614, `UserCountsPerRegionLambdaIntegrationTest` also fails, exposing the corresponding bug.

Not sure if the PR is the correct fix though. Compare the comment on the PR itself.

guozhangwang:
Thanks for the illustration @mjsax

mjsax:
Closing this PR for now -- https://issues.apache.org/jira/browse/KAFKA-3729 is reopened and this feature requires a KIP and will come later.

PR #245, opened 2019-04-23
Minor improve test stability 2

vvcephei:
A continuation of #242 . Porting integration tests to TopologyTestDriver, where applicable.

### DIFF
--- a/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
+++ b/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
@@ -24,6 +24,7 @@ import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.record.TimestampType;
 import org.apache.kafka.common.serialization.Deserializer;
+import org.apache.kafka.common.serialization.Serde;
 import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.streams.KafkaStreams;
 import org.apache.kafka.streams.KeyValue;
@@ -36,6 +37,7 @@ import org.apache.kafka.streams.state.WindowStoreIterator;
 import org.apache.kafka.test.TestUtils;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
@@ -423,5 +425,57 @@ public class IntegrationTestUtils {
       result.put(entry.getKey(), entry.getValue());
     }
     return result;
+  }
+
+  /**
+   * A Serializer/Deserializer/Serde implementation for use when you know the data is always null
+   * @param <T> The type of the stream (you can parameterize this with any type,
+   *           since we throw an exception if you attempt to use it with non-null data)
+   */
+  static class NothingSerde<T> implements Serializer<T>, Deserializer<T>, Serde<T> {
+
+    private final boolean strict;
+
+    NothingSerde(final boolean strict) {
+      this.strict = strict;
+    }
+
+    @Override
+    public void configure(final Map<String, ?> map, final boolean b) {
+
+    }
+
+    @Override
+    public T deserialize(final String s, final byte[] bytes) {
+      if (strict && bytes != null) {
+        throw new IllegalArgumentException("Expected [" + Arrays.toString(bytes) + "] to be null.");
+      } else {
+        return null;
+      }
+    }
+
+    @Override
+    public byte[] serialize(final String s, final T t) {
+      if (strict && t != null) {
+        throw new IllegalArgumentException("Expected [" + t + "] to be null.");
+      } else {
+        return null;
+      }
+    }
+
+    @Override
+    public void close() {
+
+    }
+
+    @Override
+    public Serializer<T> serializer() {
+      return this;
+    }
+
+    @Override
+    public Deserializer<T> deserializer() {
+      return this;
+    }
   }
 }
--- a/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
+++ b/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
@@ -39,6 +39,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -316,6 +317,36 @@ public class IntegrationTestUtils {
   }
 
   /**
+   * Similar to {@link IntegrationTestUtils#waitUntilMinKeyValueRecordsReceived(Properties, String, int)}, except for
+   * use with {@link TopologyTestDriver} tests. Because the test driver is synchronous, we don't need to poll for
+   * the expected number of records, and then hope that these are all the results from our test. Instead, we can
+   * just read out <em>all</em> the processing results, for use with deterministic validations.
+   *
+   * @param topic Topic to consume from
+   * @param topologyTestDriver The {@link TopologyTestDriver} to read the data records from
+   * @param keyDeserializer The {@link Deserializer} corresponding to the key type
+   * @param valueDeserializer  The {@link Deserializer} corresponding to the value type
+   * @param <K> Key type of the data records
+   * @param <V> Value type of the data records
+   * @return A {@link List} of {@link KeyValue} pairs of results from the output topic, in the order they were produced.
+   */
+  static <K, V> List<KeyValue<K, V>> drainStreamOutput(final String topic,
+                                                       final TopologyTestDriver topologyTestDriver,
+                                                       final Deserializer<K> keyDeserializer,
+                                                       final Deserializer<V> valueDeserializer) {
+    final List<KeyValue<K, V>> results = new LinkedList<>();
+    while (true) {
+      final ProducerRecord<K, V> record = topologyTestDriver.readOutput(topic, keyDeserializer, valueDeserializer);
+      if (record == null) {
+        break;
+      } else {
+        results.add(new KeyValue<>(record.key(), record.value()));
+      }
+    }
+    return results;
+  }
+
+  /**
    * Like {@link IntegrationTestUtils#produceKeyValuesSynchronously(String, Collection, Properties)}, except for use
    * with TopologyTestDriver tests, rather than "native" Kafka broker tests.
    *
@@ -393,5 +424,4 @@ public class IntegrationTestUtils {
     }
     return result;
   }
-
 }
--- a/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
+++ b/src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
@@ -434,20 +434,14 @@ public class IntegrationTestUtils {
    */
   static class NothingSerde<T> implements Serializer<T>, Deserializer<T>, Serde<T> {
 
-    private final boolean strict;
-
-    NothingSerde(final boolean strict) {
-      this.strict = strict;
-    }
-
     @Override
-    public void configure(final Map<String, ?> map, final boolean b) {
+    public void configure(final Map<String, ?> configuration, final boolean isKey) {
 
     }
 
     @Override
-    public T deserialize(final String s, final byte[] bytes) {
-      if (strict && bytes != null) {
+    public T deserialize(final String topic, final byte[] bytes) {
+      if (bytes != null) {
         throw new IllegalArgumentException("Expected [" + Arrays.toString(bytes) + "] to be null.");
       } else {
         return null;
@@ -455,9 +449,9 @@ public class IntegrationTestUtils {
     }
 
     @Override
-    public byte[] serialize(final String s, final T t) {
-      if (strict && t != null) {
-        throw new IllegalArgumentException("Expected [" + t + "] to be null.");
+    public byte[] serialize(final String topic, final T data) {
+      if (data != null) {
+        throw new IllegalArgumentException("Expected [" + data + "] to be null.");
       } else {
         return null;
       }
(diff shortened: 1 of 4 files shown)


src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -394,4 +427,55 @@ public V setValue(final V value) {
     return result;
   }
 
+  /**
+   * A Serializer/Deserializer/Serde implementation for use when you know the data is always null
+   * @param <T> The type of the stream (you can parameterize this with any type,
+   *           since we throw an exception if you attempt to use it with non-null data)
+   */
+  static class NothingSerde<T> implements Serializer<T>, Deserializer<T>, Serde<T> {
```
vvcephei: Added this because I didn't like seeing us put in a random serde, knowing that it wouldn't actually be used.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -394,4 +427,55 @@ public V setValue(final V value) {
...
+
+    private final boolean strict;
+
+    NothingSerde(final boolean strict) {
```
vvcephei: This lets you prove your assumption that it'll never be used with real data, or just drop any non-null data, depending on your test case.

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
```diff
@@ -80,7 +70,7 @@ public void shouldSumEvenNumbers() throws Exception {
     // Use a temporary directory for storing state, which will be automatically removed after the test.
     streamsConfiguration.put(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getAbsolutePath());
 
-    final KStream<Integer, Integer> input = builder.stream(inputTopic);
+    final KStream<Void, Integer> input = builder.stream(inputTopic);
```
vvcephei: Our builder isn't type-safe when you use default serdes.

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
```diff
@@ -89,33 +79,29 @@ public void shouldSumEvenNumbers() throws Exception {
...
-    final Properties producerConfig = new Properties();
-    producerConfig.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, CLUSTER.bootstrapServers());
-    producerConfig.put(ProducerConfig.ACKS_CONFIG, "all");
-    producerConfig.put(ProducerConfig.RETRIES_CONFIG, 0);
-    producerConfig.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, IntegerSerializer.class);
-    producerConfig.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, IntegerSerializer.class);
-    IntegrationTestUtils.produceValuesSynchronously(inputTopic, inputValues, producerConfig);
+    IntegrationTestUtils.produceKeyValuesSynchronously(
+      inputTopic,
+      inputValues.stream().map(i -> new KeyValue<Void, Integer>(null, i)).collect(Collectors.toList()),
```
vvcephei: if you drill into `IntegrationTestUtils.produceValuesSynchronously`, this is what we were sending before.

src/test/java/io/confluent/examples/streams/TableToTableJoinIntegrationTest.java
```diff
@@ -131,51 +116,48 @@ public void shouldJoinTwoTables() throws Exception {
         .to(outputTopic, Produced.with(Serdes.String(), Serdes.String()));
 
 
-    final KafkaStreams streams = new KafkaStreams(builder.build(), streamsConfiguration);
-    streams.start();
+    final TopologyTestDriver topologyTestDriver = new TopologyTestDriver(builder.build(), streamsConfiguration);
+
```
ableegoldman: nit: extra line here (and above?)

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -315,6 +318,36 @@ private static boolean continueConsuming(final int messagesConsumed, final int m
...
+   *
+   * @param topic Topic to consume from
+   * @param topologyTestDriver The {@link TopologyTestDriver} to read the data records from
+   * @param keyDeserializer The {@link Deserializer} corresponding to the key type
+   * @param valueDeserializer  The {@link Deserializer} corresponding to the value type
+   * @param <K> Key type of the data records
+   * @param <V> Value type of the data records
+   * @return A {@link List} of {@link KeyValue} pairs of results from the output topic, in the order they were produced.
+   */
+  static <K, V> List<KeyValue<K, V>> drainStreamOutput(final String topic,
```
jukkakarvanen: FYI; This seems to be similar than TestOutputTopic what I was proposing to add kafka-streams-test-utils in KIP-456

Please, take a look:
https://cwiki.apache.org/confluence/display/KAFKA/KIP-456%3A+Helper+classes+to+make+it+simpler+to+write+test+logic+with+TopologyTestDriver

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
```diff
@@ -89,33 +79,29 @@ public void shouldSumEvenNumbers() throws Exception {
         .reduce((v1, v2) -> v1 + v2);
     sumOfOddNumbers.toStream().to(outputTopic);
 
-    final KafkaStreams streams = new KafkaStreams(builder.build(), streamsConfiguration);
-    streams.start();
+    final TopologyTestDriver topologyTestDriver = new TopologyTestDriver(builder.build(), streamsConfiguration);
```
jukkakarvanen: I wonder would it be worth at the same time get rid of this copy paste code in this Test class and utilize the topology directly from SumLambdaExample.java

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
jukkakarvanen: Also related to this. Would it be clearer to make separate TopologyTestUtils.java class to add these methods to it instead of adding those to IntegrationTestUtils.java?

mjsax: I personally think it's ok to add it here. Keeps all test utils unified on one class. We also don't have separate classed for producer and consumer test utils.

mjsax: About the KIP. It the KIP is available, it makes sense to use the new feature. Note, that this PR is agains older branches and won't be able to use the KIP in any case.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
mjsax: Shouldn't this be the only way to use it? If it's not `strict`, it seem using this Serde would not make sense? Can you give an example when it make sense to use it non strict?

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -394,4 +427,55 @@ public V setValue(final V value) {
...
+  static class NothingSerde<T> implements Serializer<T>, Deserializer<T>, Serde<T> {
+
+    private final boolean strict;
+
+    NothingSerde(final boolean strict) {
+      this.strict = strict;
+    }
+
+    @Override
+    public void configure(final Map<String, ?> map, final boolean b) {
```
mjsax: nit: rename `b` -- should be `isKey` ?

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -394,4 +427,55 @@ public V setValue(final V value) {
...
+
+    }
+
+    @Override
+    public T deserialize(final String s, final byte[] bytes) {
```
mjsax: nit: rename `s` -- should be `topic` ?

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -394,4 +427,55 @@ public V setValue(final V value) {
...
+      if (strict && bytes != null) {
+        throw new IllegalArgumentException("Expected [" + Arrays.toString(bytes) + "] to be null.");
+      } else {
+        return null;
+      }
+    }
+
+    @Override
+    public byte[] serialize(final String s, final T t) {
```
mjsax: As above. Same for `t`

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
mjsax: Would `Object` be a better type instead of `Void` ?

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
```diff
@@ -89,33 +79,29 @@ public void shouldSumEvenNumbers() throws Exception {
...
-        outputTopic, expectedValues.size());
-    streams.close();
-    assertThat(actualValues).isEqualTo(expectedValues);
+    final Collection<Integer> actualValues = IntegrationTestUtils.drainTableOutput(
+      outputTopic,
+      topologyTestDriver,
+      new IntegerDeserializer(),
+      new IntegerDeserializer()
+    ).values();
+    assertThat(actualValues).containsExactlyElementsOf(expectedValues);
```
mjsax: Why no `isEquals`?

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: Good feedback. I did have one use case, but I've refactored it out. So, now this parameter is just tech debt. I'll remove it.

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
vvcephei: Object would be more relaxed. The idea here is to document that the keys are always `null` at this point. (`null`) is the only possible "value" of a `Void` reference.

vvcephei: But it's just documentation, since this type is never checked.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: Hi, @jukkakarvanen ! I haven't gotten a chance to look at your KIP yet, but I really like it. I'd be happy to come back and use new test-util improvements to this repo once they're available. Really, this PR is what we should have done when we added TopologyTestDriver, for example.

For now, though, I think I'd agree with @mjsax and rather leave all the test utility methods in one "bucket" to help us discover them as we maintain these tests in the future. Since TopologyTestDriver tests are also "integration tests", it seems like the name "IntegrationTestUtils" is still appropriate.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: Oops! Autogenerated code...

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
vvcephei: I'll try it again. I think I tried it, but it might have been `is()`. It failed because the expected/actual collections had different concrete types.

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
vvcephei: Oh, wow. I didn't even notice that the whole topology is just duplicated. Kind of embarrassing, actually. Thanks for the catch!

vvcephei: So, I did this and learned that the test was summing *even* numbers, while the example was summing *odd* numbers. Oops!

It's fixed now.

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
vvcephei: Weird. It worked with no problem. Thanks for the catch.

src/test/java/io/confluent/examples/streams/SumLambdaIntegrationTest.java
mjsax: Ack. I am fine either way. Was just a thought.

vvcephei:
Following @jukkakarvanen 's advice, I re-used the topology from the `SumLambdaExample` in the integration test, which switched the application from an even-sum to an odd-sum. Since I was changing the expected results anyway, I decided to showcase the fact that you can deterministically observe every transition in the output results when you're using TopologyTestDriver.

vvcephei:
Test failure was a flaky test (maybe this one will be next for porting):
`io.confluent.examples.streams.StreamToTableJoinScalaIntegrationTest.shouldCountClicksPerRegion`

Retest this, please.

vvcephei:
Oh, right, "retest" doesn't work... Kicked it off manually: https://jenkins.confluent.io/job/confluentinc-pr/job/kafka-streams-examples/job/PR-245/3/

mjsax:
No need to add a link if you trigger retesting. Just hit the button. It will should up automatically.

jukkakarvanen:
@mjsax: "I personally think it's ok to add it here. Keeps all test utils unified on one class. We also don't have separate classed for producer and consumer test utils."

No problem, but my point was both producer and consumer test utils are used when testing with EmbeddedKafka, but you never mix these with TopologyTestDriver based tests.

vvcephei:
@jukkakarvanen , I think I see what you're getting at. How about I keep it as a "todo" to clean up and re-organize the test utils at the end of this process? I have a feeling that some of the methods will become unused as I port more tests over to TTD.

vvcephei:
downstream merge complete

PR #246, opened 2019-04-25
MINOR: port more tests to TopologyTestDriver

vvcephei:
Continuing on the work of #242 and #245 , port four more tests.

PR #248, opened 2019-04-26
MINOR: port more tests to TopologyTestDriver

vvcephei:
Continue the work of #242 , #245 , and #246 by porting the rest of the integration tests that don't use Schema Registry.

The Schema Registry tests will get ported in a later PR.

### DIFF
--- a/src/main/scala/io/confluent/examples/streams/algebird/ProbabilisticCounter.scala
+++ b/src/main/scala/io/confluent/examples/streams/algebird/ProbabilisticCounter.scala
@@ -9,7 +9,7 @@ import org.apache.kafka.streams.processor.ProcessorContext
   * estimate.
   */
 class ProbabilisticCounter(val cmsStoreName: String)
-    extends Transformer[Array[Byte], String, KeyValue[String, Long]] {
+    extends Transformer[String, String, KeyValue[String, Long]] {
 
   private var cmsState: CMSStore[String] = _
   private var processorContext: ProcessorContext = _
@@ -19,7 +19,7 @@ class ProbabilisticCounter(val cmsStoreName: String)
     cmsState = this.processorContext.getStateStore(cmsStoreName).asInstanceOf[CMSStore[String]]
   }
 
-  override def transform(key: Array[Byte], value: String): KeyValue[String, Long] = {
+  override def transform(key: String, value: String): KeyValue[String, Long] = {
     // Count the record value, think: "+ 1"
     cmsState.put(value, this.processorContext.timestamp())
 
--- a/src/test/java/io/confluent/examples/streams/EventDeduplicationLambdaIntegrationTest.java
+++ b/src/test/java/io/confluent/examples/streams/EventDeduplicationLambdaIntegrationTest.java
@@ -15,18 +15,13 @@
  */
 package io.confluent.examples.streams;
 
-import io.confluent.examples.streams.kafka.EmbeddedSingleNodeKafkaCluster;
-import org.apache.kafka.clients.consumer.ConsumerConfig;
-import org.apache.kafka.clients.producer.ProducerConfig;
-import org.apache.kafka.common.serialization.ByteArrayDeserializer;
-import org.apache.kafka.common.serialization.ByteArraySerializer;
 import org.apache.kafka.common.serialization.Serdes;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.kafka.common.serialization.StringSerializer;
-import org.apache.kafka.streams.KafkaStreams;
 import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.StreamsBuilder;
 import org.apache.kafka.streams.StreamsConfig;
+import org.apache.kafka.streams.TopologyTestDriver;
 import org.apache.kafka.streams.kstream.KStream;
 import org.apache.kafka.streams.kstream.KeyValueMapper;
 import org.apache.kafka.streams.kstream.Transformer;
@@ -36,8 +31,6 @@ import org.apache.kafka.streams.state.Stores;
 import org.apache.kafka.streams.state.WindowStore;
 import org.apache.kafka.streams.state.WindowStoreIterator;
 import org.apache.kafka.test.TestUtils;
-import org.junit.BeforeClass;
-import org.junit.ClassRule;
 import org.junit.Test;
 
 import java.util.Arrays;
@@ -45,56 +38,45 @@ import java.util.List;
 import java.util.Properties;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * End-to-end integration test that demonstrates how to remove duplicate records from an input
  * stream.
- *
+ * <p>
  * Here, a stateful {@link org.apache.kafka.streams.kstream.Transformer} (from the Processor API)
  * detects and discards duplicate input records based on an "event id" that is embedded in each
  * input record.  This transformer is then included in a topology defined via the DSL.
- *
+ * <p>
  * In this simplified example, the values of input records represent the event ID by which
  * duplicates will be detected.  In practice, record values would typically be a more complex data
  * structure, with perhaps one of the fields being such an event ID.  De-duplication by an event ID
  * is but one example of how to perform de-duplication in general.  The code example below can be
  * adapted to other de-duplication approaches.
- *
+ * <p>
  * IMPORTANT:  Kafka including its Streams API support exactly-once semantics since version 0.11.
  * With this feature available, most use cases will no longer need to worry about duplicate messages
  * or duplicate processing.  That said, there will still be some use cases where you have your own
  * business rules that define when two events are considered to be "the same" and need to be
  * de-duplicated (e.g. two events having the same payload but different timestamps).  The example
  * below demonstrates how to implement your own business rules for event de-duplication.
- *
+ * <p>
  * Note: This example uses lambda expressions and thus works with Java 8+ only.
  */
 public class EventDeduplicationLambdaIntegrationTest {
 
-  @ClassRule
-  public static final EmbeddedSingleNodeKafkaCluster CLUSTER = new EmbeddedSingleNodeKafkaCluster();
-
-  private static String inputTopic = "inputTopic";
-  private static String outputTopic = "outputTopic";
-
-  private static String storeName = "eventId-store";
-
-  @BeforeClass
-  public static void startKafkaCluster() throws Exception {
-    CLUSTER.createTopic(inputTopic);
-    CLUSTER.createTopic(outputTopic);
-  }
+  private static final String storeName = "eventId-store";
 
   /**
    * Discards duplicate records from the input stream.
-   *
+   * <p>
    * Duplicate records are detected based on an event ID;  in this simplified example, the record
    * value is the event ID.  The transformer remembers known event IDs in an associated window state
    * store, which automatically purges/expires event IDs from the store after a certain amount of
    * time has passed to prevent the store from growing indefinitely.
-   *
+   * <p>
    * Note: This code is for demonstration purposes and was not tested for production usage.
    */
   private static class DeduplicationTransformer<K, V, E> implements Transformer<K, V, KeyValue<K, V>> {
@@ -118,9 +100,9 @@ public class EventDeduplicationLambdaIntegrationTest {
      *                                     ID), during the time of which any incoming duplicates of
      *                                     the event will be dropped, thereby de-duplicating the
      *                                     input.
-     * @param idExtractor extracts a unique identifier from a record by which we de-duplicate input
-     *                    records; if it returns null, the record will not be considered for
-     *                    de-duping but forwarded as-is.
+     * @param idExtractor                  extracts a unique identifier from a record by which we de-duplicate input
+     *                                     records; if it returns null, the record will not be considered for
+     *                                     de-duping but forwarded as-is.
      */
     DeduplicationTransformer(final long maintainDurationPerEventInMs, final KeyValueMapper<K, V, E> idExtractor) {
       if (maintainDurationPerEventInMs < 1) {
@@ -158,9 +140,9 @@ public class EventDeduplicationLambdaIntegrationTest {
     private boolean isDuplicate(final E eventId) {
       final long eventTime = context.timestamp();
       final WindowStoreIterator<Long> timeIterator = eventIdStore.fetch(
-          eventId,
-          eventTime - leftDurationMs,
-          eventTime + rightDurationMs);
+        eventId,
+        eventTime - leftDurationMs,
+        eventTime + rightDurationMs);
       final boolean isDuplicate = timeIterator.hasNext();
       timeIterator.close();
       return isDuplicate;
@@ -189,12 +171,12 @@ public class EventDeduplicationLambdaIntegrationTest {
   }
 
   @Test
-  public void shouldRemoveDuplicatesFromTheInput() throws Exception {
+  public void shouldRemoveDuplicatesFromTheInput() {
     final String firstId = UUID.randomUUID().toString(); // e.g. "4ff3cb44-abcb-46e3-8f9a-afb7cc74fbb8"
     final String secondId = UUID.randomUUID().toString();
     final String thirdId = UUID.randomUUID().toString();
     final List<String> inputValues = Arrays.asList(firstId, secondId, firstId, firstId, secondId, thirdId,
-        thirdId, firstId, secondId);
+      thirdId, firstId, secondId);
     final List<String> expectedValues = Arrays.asList(firstId, secondId, thirdId);
 
     //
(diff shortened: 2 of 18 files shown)


vvcephei:
Trying to compress this process, since we're in the groove of reviewing these tests now. I've ported all the remaining tests, except the SR ones.

I realized this time around that I've been forgetting to remove unnecessary configs, and also forgetting to close the TopologyTestDriver at the end of the tests, so I've gone back and patched the tests I'd already written.

mjsax:
\cc @miguno

miguno:
@vvcephei : I have an in-progress PR (https://github.com/confluentinc/kafka-streams-examples/pull/235) that adds two new examples, with implementations for Java and Scala, to this repo. I have already begun refactoring them to use the Topo driver, but I'd prefer to pause the PR until you are done with these changes so that I am not duplicating code, get merge conflicts, etc.

Would you mind pinging me once you feel that the test migration (to use topo driver) is completed?  Then I can pick up "my" PR again.

vvcephei:
@miguno , sure! Sorry for stirring the pot, but hopefully these changes make maintaining these examples a lot easier.

src/main/scala/io/confluent/examples/streams/algebird/ProbabilisticCounter.scala
```diff
@@ -9,7 +9,7 @@ import org.apache.kafka.streams.processor.ProcessorContext
   * estimate.
   */
 class ProbabilisticCounter(val cmsStoreName: String)
-    extends Transformer[Array[Byte], String, KeyValue[String, Long]] {
```
vvcephei: I'm not sure why, but these changes were necessary because IDEA reports scala compile errors, even when the code actually compiles. Even though this is really a bug in IDEA, I figured that it's preferable to present code that doesn't have apparent errors in such a common IDE.

miguno: What does IDEA complain about?

miguno: This should be necessary because, further down below, you made this change:

From

```scala
override def transform(key: Array[Byte], value: String): KeyValue[String, Long] = {
```
to 
```scala
override def transform(key: String, value: String): KeyValue[String, Long] = {
```
Note the type change for `key`.

src/test/scala/io/confluent/examples/streams/ProbabilisticCountingScalaIntegrationTest.scala
```diff
@@ -82,13 +71,9 @@ class ProbabilisticCountingScalaIntegrationTest extends AssertionsForJUnit {
     val streamsConfiguration: Properties = {
       val p = new Properties()
       p.put(StreamsConfig.APPLICATION_ID_CONFIG, "probabilistic-counting-scala-integration-test")
-      p.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, cluster.bootstrapServers())
+      p.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config")
       p.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.ByteArray.getClass.getName)
```
miguno: @vvcephei : Do we even need to set the default serdes anymore?  In the case of the Scala API for KStreams, the API should automatically pick up the right serdes via Scala implicits (at least for the primitive types like String etc., which are all supported OOTB by KStreams in Scala).

src/test/scala/io/confluent/examples/streams/ProbabilisticCountingScalaIntegrationTest.scala
```diff
@@ -99,63 +84,52 @@ class ProbabilisticCountingScalaIntegrationTest extends AssertionsForJUnit {
...
-          cmsStoreName)
+      .flatMapValues(textLine => textLine.toLowerCase.split("\\W+").toIterable.asJava)
+      .transform(
+        new TransformerSupplier[String, String, KeyValue[String, Long]] {
+          override def get() = new ProbabilisticCounter(cmsStoreName)
+        },
+        cmsStoreName)
 
     // Trick to re-use Kafka's serde for java.lang.Long for scala.Long.
     val longSerde: Serde[Long] = Serdes.Long().asInstanceOf[Serde[Long]]
```
miguno: Unrelated note: We should no longer need this trick.  With KStreams native Scala API, we don't need to set serdes explicitly anymore.

src/test/scala/io/confluent/examples/streams/ProbabilisticCountingScalaIntegrationTest.scala
```diff
@@ -99,63 +84,52 @@ class ProbabilisticCountingScalaIntegrationTest extends AssertionsForJUnit {
...
     // Write the results back to Kafka.
     approximateWordCounts.to(outputTopic, Produced.`with`(Serdes.String(), longSerde))
```
miguno: For Scala we should no longer need to use `Produced.with` as well as `Consumed.with`().  Just `stream(topic)`, `table(topic)` and `to(topic)` will suffice nowadays.

src/main/scala/io/confluent/examples/streams/algebird/ProbabilisticCounter.scala
vvcephei: Yep, it was saying there was a type mismatch somewhere in the topology, which went away (for unknown reasons) when I switched the key to `String`. Pushing that change through resulted in the change to the counter.

src/test/scala/io/confluent/examples/streams/ProbabilisticCountingScalaIntegrationTest.scala
vvcephei: Unsure! I didn't think about that. I'll try it out.

vvcephei: Ah, these examples haven't been ported to the actual scala API yet. I think I'll defer that work till a later PR, to keep this one nice and formulaic.

src/test/scala/io/confluent/examples/streams/ProbabilisticCountingScalaIntegrationTest.scala
vvcephei: Same explanation as above. This will be true once we port to the scala dsl.

src/test/scala/io/confluent/examples/streams/ProbabilisticCountingScalaIntegrationTest.scala
vvcephei: Same also as above. We need to port to the Scala DSL first.

src/test/java/io/confluent/examples/streams/WordCountLambdaIntegrationTest.java
```diff
@@ -15,79 +15,69 @@
...
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
+import static io.confluent.examples.streams.IntegrationTestUtils.mkEntry;
+import static io.confluent.examples.streams.IntegrationTestUtils.mkMap;
```
abbccdda: Could we use `mkMap` in `code/kafka/clients/src/main/java/org/apache/kafka/common/utils/Utils.java` instead? Potential we could avoid double implementing the same feature.

src/test/java/io/confluent/examples/streams/UserCountsPerRegionLambdaIntegrationTest.java
```diff
@@ -116,29 +111,33 @@ public void shouldCountUsersPerRegion() {
...
-      topologyTestDriver,
-      new StringDeserializer(),
-      new LongDeserializer()
-    );
-    assertThat(actualClicksPerRegion).isEqualTo(expectedUsersPerRegion);
+    try {
+      //
+      // Step 2: Publish user-region information.
+      //
+
```
abbccdda: Any specific reason for this spacing in last version? Seems unnecessary.

src/test/java/io/confluent/examples/streams/TableToTableJoinIntegrationTest.java
```diff
@@ -110,53 +103,57 @@ public void shouldJoinTwoTables() {
...
-    assertThat(keyValueIterator).containsExactlyElementsOf(expectedResultsForJoinStateStore);
+      final KeyValueIterator<String, String> keyValueIterator = readOnlyKeyValueStore.all();
+      assertThat(keyValueIterator).containsExactlyElementsOf(expectedResultsForJoinStateStore);
 
-    assertThat(actualResults).containsExactlyElementsOf(expectedResults);
+      assertThat(actualResults).containsExactlyElementsOf(expectedResults);
+    } finally {
+      topologyTestDriver.close();
+    }
   }
```
abbccdda: This line could be removed.

src/test/java/io/confluent/examples/streams/StreamToStreamJoinIntegrationTest.java
```diff
@@ -104,48 +99,52 @@ public void shouldJoinTwoStreams() throws Exception {
...
+      // Step 4: Verify the application's output data.
+      //
+      final List<KeyValue<String, String>> actualResults =
+        IntegrationTestUtils.drainStreamOutput(
+          outputTopic,
+          topologyTestDriver,
+          new StringDeserializer(),
+          new StringDeserializer()
+        );
+      assertThat(actualResults).containsExactlyElementsOf(expectedResults);
```
abbccdda: Seems we refactored some example to use map instead of key-value list. Do we want to do this for all similar use cases?

src/test/java/io/confluent/examples/streams/WordCountLambdaIntegrationTest.java
vvcephei: Huh, I thought that was an internal class. It certainly should be. This means we should have had a KIP to propose the mkMap method for Utils.

More generally, it's not good "hygine" to depend on another project's utils, even if they accidentally fail to keep them private. Kafka Client's "job" is to provide the client libraries, not to provide a way to make maps with varargs, so we shouldn't expect this to be a supported API.

Especially for a case like this, where the logic is so trivial, we should strongly prefer to copy the code, rather than introduce another external dependency. Every line of code is a maintenance burden, but every external dependency is like 10 maintenance burdens.

src/test/java/io/confluent/examples/streams/UserCountsPerRegionLambdaIntegrationTest.java
vvcephei: probably just to create more visual gap to enhance readability.

src/test/java/io/confluent/examples/streams/TableToTableJoinIntegrationTest.java
vvcephei:
```suggestion
```

src/test/java/io/confluent/examples/streams/StreamToStreamJoinIntegrationTest.java
vvcephei: no, only for the tabular examples. In this case, we're producing a stream, so we're verifying the stream of updates is as expected.

src/test/java/io/confluent/examples/streams/UserCountsPerRegionLambdaIntegrationTest.java
vvcephei: Note, the spacing didn't change in this PR. But I can remove the blank line if it bothers you.

```suggestion
```

src/test/java/io/confluent/examples/streams/FanoutLambdaIntegrationTest.java
```diff
@@ -88,56 +69,56 @@ public void shouldFanoutTheInput() throws Exception {
...
     streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
-    streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
 
+    final String inputTopicA = "A";
+    final String outputTopicB = "B";
+    final String outputTopicC = "C";
     final KStream<byte[], String> stream1 = builder.stream(inputTopicA);
-    final KStream<byte[], String> stream2 = stream1.mapValues((v -> v.toUpperCase()));
-    final KStream<byte[], String> stream3 = stream1.mapValues(v -> v.toLowerCase());
+    final KStream<byte[], String> stream2 = stream1.mapValues(((ValueMapper<String, String>) String::toUpperCase));
```
bbejeck: nit: I understand the cast to disambiguate from`ValueMapperWithKey` but is casting any better than `v -> v.toUpperCase()`?

src/test/java/io/confluent/examples/streams/FanoutLambdaIntegrationTest.java
```diff
@@ -88,56 +69,56 @@ public void shouldFanoutTheInput() throws Exception {
...
+    final KStream<byte[], String> stream3 = stream1.mapValues((ValueMapper<String, String>) String::toLowerCase);
```
bbejeck: ditto

src/test/java/io/confluent/examples/streams/MapFunctionLambdaIntegrationTest.java
```diff
@@ -69,41 +52,42 @@ public void shouldUppercaseTheInput() throws Exception {
...
+    streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
     streamsConfiguration.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.ByteArray().getClass().getName());
     streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
-    streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
 
+    final String inputTopic = "inputTopic";
+    final String outputTopic = "outputTopic";
     final KStream<byte[], String> input = builder.stream(inputTopic);
-    final KStream<byte[], String> uppercased = input.mapValues(v -> v.toUpperCase());
+    final KStream<byte[], String> uppercased = input.mapValues((ValueMapper<String, String>) String::toUpperCase);
```
bbejeck: as above

src/test/java/io/confluent/examples/streams/MixAndMatchLambdaIntegrationTest.java
```diff
@@ -137,44 +119,47 @@ public void shouldAnonymizeTheInput() throws Exception {
...
-    streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+
+    final String inputTopic = "inputTopic";
+    final String outputTopic = "outputTopic";
 
     final KStream<byte[], String> input = builder.stream(inputTopic);
     final KStream<byte[], String> uppercasedAndAnonymized = input
-        .mapValues(v -> v.toUpperCase())
-        .transform(AnonymizeIpAddressTransformer::new);
+      .mapValues((ValueMapper<String, String>) String::toUpperCase)
```
bbejeck: as above

src/test/java/io/confluent/examples/streams/FanoutLambdaIntegrationTest.java
vvcephei: It's debatable... I just took the suggestion from IDEA. Happy to roll it back.

bbejeck: just leave it.

vvcephei: too late :)

src/test/java/io/confluent/examples/streams/EventDeduplicationLambdaIntegrationTest.java
```diff
@@ -36,65 +31,52 @@
...
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * End-to-end integration test that demonstrates how to remove duplicate records from an input
  * stream.
- *
+ * <p>
```
vvcephei:
```suggestion
 * <p> 
```

vvcephei:
Ok, it looks like the build might be fixed now. I'm reverting all the junk I've tried to fix it, and we'll see if we get a clean build now.

vvcephei:
The build passed! Time to take the money and run...

src/test/scala/io/confluent/examples/streams/ProbabilisticCountingScalaIntegrationTest.scala
vvcephei: @miguno , during the downstream merge, I realize this was added in a later release, so your feedback is already implemented in downstream branches.

src/test/scala/io/confluent/examples/streams/ProbabilisticCountingScalaIntegrationTest.scala
vvcephei: @miguno , during the downstream merge, I realize this was added in a later release, so your feedback is already implemented in downstream branches.

vvcephei:
downstream merge complete.

PR #250, opened 2019-05-02
MINOR: Block on streams start until in a RUNNING state

bbejeck:
This PR is an attempt to help reduce the flakiness of some of the microservice tests.  From looking at the logs, it seems it may take a streams app some time to get into a running state, so the idea is to have the tests block until streams goes from `REBALANCING` to a `RUNNING` state

I'm targeting the `3.3.0-post` branch, so this will hit all versions we're responsible for keeping the tests green.

### DIFF
--- a/src/main/java/io/confluent/examples/streams/microservices/EmailService.java
+++ b/src/main/java/io/confluent/examples/streams/microservices/EmailService.java
@@ -52,11 +52,8 @@ public class EmailService implements Service {
 
     });
     streams.start();
-    final boolean rebalanceCompleted;
-
     try {
-      rebalanceCompleted = startLatch.await(60, TimeUnit.SECONDS);
-      if (!rebalanceCompleted) {
+      if (!startLatch.await(60, TimeUnit.SECONDS)) {
         throw new RuntimeException("Streams never finished rebalancing on startup");
       }
     } catch (final InterruptedException e) {
--- a/src/main/java/io/confluent/examples/streams/microservices/EmailService.java
+++ b/src/main/java/io/confluent/examples/streams/microservices/EmailService.java
@@ -11,7 +11,11 @@ import static io.confluent.examples.streams.microservices.util.MicroserviceUtils
 import io.confluent.examples.streams.avro.microservices.Customer;
 import io.confluent.examples.streams.avro.microservices.Order;
 import io.confluent.examples.streams.avro.microservices.Payment;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import org.apache.kafka.streams.KafkaStreams;
+import org.apache.kafka.streams.KafkaStreams.State;
+import org.apache.kafka.streams.KafkaStreams.StateListener;
 import org.apache.kafka.streams.kstream.GlobalKTable;
 import org.apache.kafka.streams.kstream.JoinWindows;
 import org.apache.kafka.streams.kstream.KStream;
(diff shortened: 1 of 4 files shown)


bbejeck:
\cc @benstopford

src/main/java/io/confluent/examples/streams/microservices/EmailService.java
```diff
@@ -40,7 +44,24 @@ public EmailService(final Emailer emailer) {
...
     streams.start();
+    final boolean rebalanceCompleted;
+
+    try {
+      rebalanceCompleted = startLatch.await(60, TimeUnit.SECONDS);
+      if (!rebalanceCompleted) {
+        throw new RuntimeException("Streams never finished rebalancing on startup");
+      }
+    } catch (final InterruptedException e) {
+       Thread.currentThread().interrupt();
```
bbejeck: I first attempted to change the signature of the `Service.start` method to include `throws InterruptedException` and then simply re-throw the `InterruptedException`.   But the changes were going to be more extensive than I felt this fix warranted.  So I'm opting for the next best approach.

This applies to other files as well.

ableegoldman:
Not related to this PR, but when I was messing around trying to write my own Streams app it wasn't very obvious that I had to block on the REBALANCING -> RUNNING transition. Maybe this could be better documented somewhere? Or we could have a version of streams.start() that does this for us or returns a stateListener, or something to indicate how to handle this? Maybe I was just new to Streams and it should be obvious, but I would argue for low entry barrier it should be obvious to someone just picking this up for the first time

bbejeck:
Thanks for the review @ableegoldman.  To be clear, users shouldn't have to block on streams getting to a `RUNNING` state. This PR is a workaround for the flaky tests so that we don't start the timer on the assertions for reading/writing to topics until streams are entirely functional.

abbccdda:
Could we refactor this code change into a common function?

ableegoldman:
> Thanks for the review @ableegoldman. To be clear, users shouldn't have to block on streams getting to a `RUNNING` state. This PR is a workaround for the flaky tests so that we don't start the timer on the assertions for reading/writing to topics until streams are entirely functional.

Sorry, I meant for interactive queries not in general

bbejeck:
>Sorry, I meant for interactive queries not in general
DOH! yeah, I should have figured that's what you meant!

bbejeck:
> Could we refactor this code change into a common function?

@abbccdda  the service classes don't have a common base class and IMHO adding a base class to have a common function, in this case, isn't needed.

src/main/java/io/confluent/examples/streams/microservices/EmailService.java
```diff
@@ -40,7 +44,24 @@ public EmailService(final Emailer emailer) {
...
     streams.cleanUp(); //don't do this in prod as it clears your state stores
+    final CountDownLatch startLatch = new CountDownLatch(1);
+    streams.setStateListener((newState, oldState) -> {
+      if (newState == State.RUNNING && oldState == State.REBALANCING) {
+        startLatch.countDown();
+      }
+
+    });
     streams.start();
+    final boolean rebalanceCompleted;
```
vvcephei: nit: is this variable necessary? It seems like :
```java
if (!startLatch.await(60, TimeUnit.SECONDS) {
        throw new RuntimeException("Streams never finished rebalancing on startup");
}
```

would do the same thing?

bbejeck: ack

bbejeck:
merged to all downstream branches via pint

PR #252, opened 2019-05-15
MINOR: migrate GenericAvroIntegrationTest to TTD

vvcephei:
Continue the work of #242 , #245 , #246 , and #248 by establishing a pattern for GenericRecord Avro tests to use TopologyTestDriver.

### DIFF
--- a/src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
+++ b/src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
@@ -15,7 +15,8 @@
  */
 package io.confluent.examples.streams;
 
-import io.confluent.examples.streams.kafka.EmbeddedSingleNodeKafkaCluster;
+import io.confluent.examples.streams.utils.MockGenericAvroSerde;
+import io.confluent.kafka.schemaregistry.client.MockSchemaRegistryClient;
 import io.confluent.kafka.serializers.AbstractKafkaAvroSerDeConfig;
 import io.confluent.kafka.serializers.KafkaAvroDeserializer;
 import io.confluent.kafka.serializers.KafkaAvroSerializer;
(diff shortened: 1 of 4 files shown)


src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -80,10 +77,11 @@ public void shouldRoundTripGenericAvroDataThroughKafka() throws Exception {
 
     final Properties streamsConfiguration = new Properties();
     streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "generic-avro-integration-test");
-    streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, CLUSTER.bootstrapServers());
+    streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
     streamsConfiguration.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.ByteArray().getClass().getName());
-    streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, GenericAvroSerde.class);
-    streamsConfiguration.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, CLUSTER.schemaRegistryUrl());
+    streamsConfiguration.put("mock.schema.registry.client", mockSchemaRegistryClient);
+    streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, MockGenericAvroSerde.class);
```
vvcephei: This is a sticky point. The GenericAvroSerde class (and company) are totally incompatible with providing a MockSchemaRegistryClient. Even leaving aside passing one vial the `configure` method, they have only package-private constructors that accept a pre-constructed client.

The only thing I could think to do is just re-implement the GenericAvroSerde/Serializer/Deserializer classes (which are trivial) and let them grab a mock SR client out of the config. I searched around, but didn't find any better ideas "in the wild".

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -97,46 +95,48 @@ public void shouldRoundTripGenericAvroDataThroughKafka() throws Exception {
     // However, in the code below we intentionally override the default serdes in `to()` to
     // demonstrate how you can construct and configure a generic Avro serde manually.
     final Serde<String> stringSerde = Serdes.String();
-    final Serde<GenericRecord> genericAvroSerde = new GenericAvroSerde();
+    final Serde<GenericRecord> genericAvroSerde = new MockGenericAvroSerde();
```
vvcephei: Now that we have this class, we might as well use it.

vvcephei:
Thanks, @bbejeck. I agree, and this is part of the reason I was thinking to push the serde changes upstream. Having to use different serdes in testing vs. production seems sketchy.

vvcephei:
@bbejeck Ah. This PR is targeted for the `-post` branch. I can only propose that SR change for not-yet-released versions (obviously), so I think we need to consider merging this code as-is, and then pursue an upstream change for future versions.

Does that sound good?

vvcephei:
Thanks for the review. Merging now...

PR #253, opened 2019-05-22
Add a sample of custom window

jeanlouisboudart:
Here is a basic example of a custom daily window starting every day at 6pm.

It also shows a potential problem with timezone having daylight saving time (https://en.wikipedia.org/wiki/Daylight_saving_time).

### DIFF
--- a/src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
+++ b/src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
@@ -27,7 +27,15 @@ import org.apache.kafka.streams.StreamsBuilder;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.Topology;
 import org.apache.kafka.streams.TopologyTestDriver;
-import org.apache.kafka.streams.kstream.*;
+import org.apache.kafka.streams.kstream.KStream;
+import org.apache.kafka.streams.kstream.KTable;
+import org.apache.kafka.streams.kstream.Materialized;
+import org.apache.kafka.streams.kstream.Printed;
+import org.apache.kafka.streams.kstream.Produced;
+import org.apache.kafka.streams.kstream.Suppressed;
+import org.apache.kafka.streams.kstream.TimeWindowedDeserializer;
+import org.apache.kafka.streams.kstream.Windowed;
+import org.apache.kafka.streams.kstream.WindowedSerdes;
 import org.apache.kafka.streams.kstream.internals.TimeWindow;
 import org.apache.kafka.streams.test.ConsumerRecordFactory;
 import org.apache.kafka.streams.test.OutputVerifier;
@@ -38,7 +46,9 @@ import java.time.Duration;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
-import java.util.*;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Properties;
 import java.util.stream.Collectors;
 
 
--- a/src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
+++ b/src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
@@ -155,7 +155,7 @@ public class CustomWindowTest {
         //This test illustrate problems with daylight savings
         //Some timezone have daylight savings time (DST) resulting in two days in year that have either 23 or 25 hours.
         //Kafka streams currently support only fixed period for the moment.
-        ZoneId zoneWithDST = ZoneId.of("Europe/Paris");
+        final ZoneId zoneWithDST = ZoneId.of("Europe/Paris");
 
         final List<MyEvent> inputValues = Arrays.asList(
                 new MyEvent(1, ZonedDateTime.of(2019, 3, 30, 1, 39, 0, 0, zoneWithDST)),
@@ -184,7 +184,7 @@ public class CustomWindowTest {
         verify(inputValues, expectedValues, zoneWithDST);
     }
 
-    private void verify(List<MyEvent> inputValues, List<ExpectedResult> expectedValues, ZoneId zoneId) {
+    private void verify(final List<MyEvent> inputValues, final List<ExpectedResult> expectedValues, final ZoneId zoneId) {
 
         final Properties streamsConfiguration = new Properties();
         streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "sum-lambda-integration-test");
@@ -198,9 +198,9 @@ public class CustomWindowTest {
         // Use a temporary directory for storing state, which will be automatically removed after the test.
         streamsConfiguration.put(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getAbsolutePath());
 
-        Topology topology = buildKafkaStreamTopology(zoneId);
+        final Topology topology = buildKafkaStreamTopology(zoneId);
 
-        TopologyTestDriver testDriver = new TopologyTestDriver(topology, streamsConfiguration);
+        final TopologyTestDriver testDriver = new TopologyTestDriver(topology, streamsConfiguration);
 
         injectFakeData(inputValues, testDriver);
         verifyResults(expectedValues, testDriver);
@@ -208,7 +208,7 @@ public class CustomWindowTest {
         testDriver.close();
     }
 
-    private Topology buildKafkaStreamTopology(ZoneId zoneId) {
+    private Topology buildKafkaStreamTopology(final ZoneId zoneId) {
         final StreamsBuilder builder = new StreamsBuilder();
 
         final KStream<String, Integer> input = builder.stream(inputTopic);
@@ -231,7 +231,7 @@ public class CustomWindowTest {
         Integer value;
         ZonedDateTime timestamp;
 
-        public MyEvent(Integer value, ZonedDateTime timestamp) {
+        public MyEvent(final Integer value, final ZonedDateTime timestamp) {
             this.value = value;
             this.timestamp = timestamp;
         }
@@ -243,7 +243,7 @@ public class CustomWindowTest {
         ZonedDateTime start;
         ZonedDateTime end;
 
-        public ExpectedResult(Integer key, Integer value, ZonedDateTime start, ZonedDateTime end) {
+        public ExpectedResult(final Integer key, final Integer value, final ZonedDateTime start, final ZonedDateTime end) {
             this.key = key;
             this.value = value;
             this.start = start;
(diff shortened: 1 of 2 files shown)


src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
```diff
@@ -0,0 +1,254 @@
...
+    private static final String outputTopic = "outputTopic";
+    private static final ZoneId zone = ZoneOffset.UTC;
+
+    @Test
+    public void shouldSumNumbersOnSameDay() {
+        final List<MyEvent> inputValues = Arrays.asList(
+                new MyEvent(1, ZonedDateTime.of(2019, 1, 1, 16, 29, 0, 0, zone)),
+                new MyEvent(2, ZonedDateTime.of(2019, 1, 1, 16, 30, 0, 0, zone)),
+                new MyEvent(7, ZonedDateTime.of(2019, 1, 1, 16, 31, 0, 0, zone)),
+                new MyEvent(7, ZonedDateTime.now())
```
vvcephei: Is this last event to flush everything through suppression? Might be clearer to explicitly provide a record that's at time `window end + grace period` for this purpose.

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
```diff
@@ -0,0 +1,254 @@
...
+                        ZonedDateTime.of(2019, 3, 29, 18, 0, 0, 0, zoneWithDST),
+                        ZonedDateTime.of(2019, 3, 30, 18, 0, 0, 0, zoneWithDST)),
+                new ExpectedResult(1,
+                        10,
+                        ZonedDateTime.of(2019, 3, 30, 18, 0, 0, 0, zoneWithDST),
+                        //This get one extra hour due to time shift in Daylight saving
+                        //As a user i would expect it to end on 31th at 6pm.
+
+                        //The limitation seems to come from TimeWindowSerializer / TimeWindowDeserializer as we serialize only start date.
+                        //Suggestion: By serializing both start and end of window, we could support more complex cases with non fixed time window and address daylight savings on daily windows.
```
vvcephei: Yes, this is exactly right. Feel free to open a ticket for this, and thanks for providing an example of how to accomplish custom time windows that explicitly includes this edge case!

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+import java.time.Duration;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.*;
+import java.util.stream.Collectors;
+
+
+public class CustomWindowTest {
```
vvcephei: These tests are very well structured. Thanks!

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
vvcephei: Also, it's worth noting that days will have different lengths under other conditions than just leap year. Leap seconds, for example.

This is actually part of the reason we use "milliseconds since the epoch" as our clock, and align our time windows to some interval on that clock. That way, the internal clock is both monotonic and regular, neither of which are properties that real-life clocks possess.

Of course, this is an academic choice, since it simplifies the implementation at the expense of realistic use cases like "aggregate daily at midnight".

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
```diff
@@ -0,0 +1,254 @@
...
+                .groupByKey()
+                .windowedBy(new DailyTimeWindows(zoneId, Duration.ofMinutes(30)))
+                // A simple sum of value
+                .reduce((v1, v2) -> v1 + v2, Materialized.with(Serdes.Integer(), Serdes.Integer()))
+                // We only care about final result
+                .suppress(Suppressed.untilWindowCloses(Suppressed.BufferConfig.unbounded()));
+        sumOfOddNumbers.toStream().print(Printed.toSysOut());
+        sumOfOddNumbers.toStream().to(outputTopic, Produced.with(WindowedSerdes.timeWindowedSerdeFrom(Integer.class), Serdes.Integer()));
+
+        TopologyTestDriver testDriver = new TopologyTestDriver(builder.build(), streamsConfiguration);
```
vvcephei: Oh, you should be able to use try-with-resources here.

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
```diff
@@ -0,0 +1,254 @@
...
+
+    private void verifyResults(List<ExpectedResult> expectedValues, TopologyTestDriver testDriver) {
+        for (ExpectedResult expectedValue : expectedValues) {
+            ProducerRecord<Windowed<Integer>, Integer> results = testDriver.readOutput(outputTopic, new TimeWindowedDeserializer(new IntegerDeserializer(), Duration.ofDays(1).toMillis()), new IntegerDeserializer());
+            OutputVerifier.compareKeyValue(results, expectedValue.toWindowed(), expectedValue.value);
+
+        }
+    }
+
+    private void injectFakeData(List<MyEvent> inputValues, TopologyTestDriver testDriver) {
```
vvcephei: There are now some utilities in IntegrationTestUtils that you could use for this and for `verifyResults`.

jeanlouisboudart: I tried but, it doesn't support providing event with given timestamp yet :)

src/test/java/io/confluent/examples/streams/window/DailyTimeWindows.java
```diff
@@ -0,0 +1,76 @@
...
+        this.zoneId = zoneId;
+        this.grace = grace.toMillis();
+    }
+
+    @Override
+    public Map<Long, TimeWindow> windowsFor(long timestamp) {
+        Instant instant = Instant.ofEpochMilli(timestamp);
+
+        ZonedDateTime zonedDateTime = instant.atZone(zoneId);
+        ZonedDateTime startTime = zonedDateTime.getHour() >= 18 ? zonedDateTime.truncatedTo(ChronoUnit.DAYS).withHour(18) : zonedDateTime.truncatedTo(ChronoUnit.DAYS).minusDays(1).withHour(18);
```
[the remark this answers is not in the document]
jeanlouisboudart: done :)

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
```diff
@@ -0,0 +1,254 @@
...
+                        //As a user i would expect it to end on 31th at 6pm.
+
+                        //The limitation seems to come from TimeWindowSerializer / TimeWindowDeserializer as we serialize only start date.
+                        //Suggestion: By serializing both start and end of window, we could support more complex cases with non fixed time window and address daylight savings on daily windows.
+                        ZonedDateTime.of(2019, 3, 31, 19, 0, 0, 0, zoneWithDST))
+        );
+        kafkaStreamsMagic(inputValues, expectedValues, zoneWithDST);
+    }
+
+    private void kafkaStreamsMagic(List<MyEvent> inputValues, List<ExpectedResult> expectedValues, ZoneId zoneId) {
```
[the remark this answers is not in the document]
jeanlouisboudart: done

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
jeanlouisboudart: done

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
```diff
@@ -0,0 +1,270 @@
...
+import org.apache.kafka.streams.test.ConsumerRecordFactory;
+import org.apache.kafka.streams.test.OutputVerifier;
+import org.apache.kafka.test.TestUtils;
+import org.junit.Test;
+
+import java.time.Duration;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.*;
```
ableegoldman: nit: Do we allow wildcard imports here?

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
```diff
@@ -0,0 +1,270 @@
...
+import org.apache.kafka.common.header.internals.RecordHeaders;
+import org.apache.kafka.common.serialization.IntegerDeserializer;
+import org.apache.kafka.common.serialization.IntegerSerializer;
+import org.apache.kafka.common.serialization.Serdes;
+import org.apache.kafka.common.serialization.StringSerializer;
+import org.apache.kafka.streams.StreamsBuilder;
+import org.apache.kafka.streams.StreamsConfig;
+import org.apache.kafka.streams.Topology;
+import org.apache.kafka.streams.TopologyTestDriver;
+import org.apache.kafka.streams.kstream.*;
```
ableegoldman: same

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
jeanlouisboudart: I removed star import

src/test/java/io/confluent/examples/streams/window/CustomWindowTest.java
jeanlouisboudart: I removed star import

vvcephei:
Looks like the build failed with a linter check that all variables are declared `final`. @jeanlouisboudart , if you want to get a quick fix for this (and if you're using IDEA or another IDE), you can enable a code validation in your IDE to warn you when variables could be final. Then, it'll prompt you for all these locations, and also let you click a button to fix them.

vvcephei:
I've merged to `5.2.1-post`, and done the downstream merge to the later branches. `master` is currently broken until a 5.4.0-SNAPSHOT is published for our dependencies, but I'm running a final verification on `5.3.x` to make sure the downstream merge didn't break anything.

Thanks for the contribution, @jeanlouisboudart !

src/test/java/io/confluent/examples/streams/window/DailyTimeWindows.java
```diff
@@ -0,0 +1,78 @@
...
+    public Map<Long, TimeWindow> windowsFor(final long timestamp) {
+        final Instant instant = Instant.ofEpochMilli(timestamp);
+
+        final ZonedDateTime zonedDateTime = instant.atZone(zoneId);
+        final ZonedDateTime startTime = zonedDateTime.getHour() >= startHour ? zonedDateTime.truncatedTo(ChronoUnit.DAYS).withHour(startHour) : zonedDateTime.truncatedTo(ChronoUnit.DAYS).minusDays(1).withHour(startHour);
+        final ZonedDateTime endTime = startTime.plusDays(1);
+
+        final Map<Long, TimeWindow> windows = new LinkedHashMap<>();
+        windows.put(toEpochMilli(startTime), new TimeWindow(toEpochMilli(startTime), toEpochMilli(endTime)));
+        return windows;
```
peterjurkovic: Does it make sense to cache a result of this method instead of creating new LinkedHashMap + TimeWindow on every invocation?

mjsax: That might be possible. Not sure how much you gain though. I would only consider it, if you know it's a performance issue. The question is also, how many windows do you cache?  While time progressed, the set of "hot" windows changes.

PR #258, opened 2019-06-01
MINOR: Change set vs seq for java converters

bbejeck:
The `KafkaZkClient` now returns a `Set` instead of a `Seq` for `getAllTopicsInCluster` so we needed to switch to `JavaConverters.setAsJavaSetConverter` method

### DIFF
--- a/src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
+++ b/src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
@@ -253,6 +253,9 @@ public class EmbeddedSingleNodeKafkaCluster extends ExternalResource {
     @Override
     public boolean conditionMet() {
       //TODO once KAFKA-6098 is fixed use AdminClient to verify topics have been deleted
+
+      // The Set returned from JavaConverters.setAsJavaSetConverter does not support the remove
+      // method so we need to continue to wrap in a HashSet
       final Set<String> allTopics = new HashSet<>(
           JavaConverters.setAsJavaSetConverter(broker.kafkaServer().zkClient().getAllTopicsInCluster()).asJava());
       return !allTopics.removeAll(deletedTopics);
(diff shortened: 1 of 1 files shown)


src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
```diff
@@ -254,7 +254,7 @@ private TopicsDeletedCondition(final String... topics) {
     public boolean conditionMet() {
       //TODO once KAFKA-6098 is fixed use AdminClient to verify topics have been deleted
       final Set<String> allTopics = new HashSet<>(
-          JavaConverters.seqAsJavaListConverter(broker.kafkaServer().zkClient().getAllTopicsInCluster()).asJava());
+          JavaConverters.setAsJavaSetConverter(broker.kafkaServer().zkClient().getAllTopicsInCluster()).asJava());
```
mjsax: If we get a `Set` now, do we still need to wrap it in a `HashSet`? I think we only needed the wrapping `HashSet` to convert the `List`?

bbejeck: ack

bbejeck: Actually, we should keep it as it is.

 `JavaConverters.setAsJavaSetConverter(x).asJava`  returns a  wrapped `Set` which does not support `remove`
```
java.lang.UnsupportedOperationException: remove
 at scala.collection.convert.Wrappers$SetWrapper$$anon$4.remove(Wrappers.scala:123)
 at java.util.AbstractCollection.remove(AbstractCollection.java:293)
 at java.util.AbstractSet.removeAll(AbstractSet.java:174)
 at io.confluent.examples.streams.kafka.EmbeddedSingleNodeKafkaCluster$TopicsDeletedCondition.conditionMet(EmbeddedSingleNodeKafkaCluster.java:258)
```
I'd prefer just to keep as is.  If you insist I can make the change

mjsax: I see. Fine with me to keep as is -- maybe add a comment why we copy the given set into a `HashSet` ?

bbejeck:
updated this

src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
bbejeck: Ack, can I get a +1 so we can merge?

PR #260, opened 2019-06-10
Use upstream mock SR support

vvcephei:
* removes the (hacky) mock schema registry serdes
* demonstrate use of upstream support for one of each kind of SR usage in the examples


vvcephei:
Note! This PR depends on upstream code that I haven't submitted pull requests yet. I'd like to get confirmation that this API looks good before submitting the upstream PRs.

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -51,18 +50,28 @@
  */
 public class GenericAvroIntegrationTest {
```
vvcephei: This example now just demonstrates how to test with the mock SR.

src/test/java/io/confluent/examples/streams/SessionWindowsExampleTest.java
```diff
@@ -16,57 +16,57 @@
...
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
 
 public class SessionWindowsExampleTest {
```
vvcephei: We couldn't migrate this example before because of the dependency on SR. Now we can!

This demonstrates a case where we can entirely move away from the embedded Kafka, as we replace KafkaStreams with TTD and we mock the SR.

src/test/java/io/confluent/examples/streams/interactivequeries/kafkamusic/KafkaMusicExampleTest.java
```diff
@@ -71,13 +74,24 @@
 
   @ClassRule
   public static final EmbeddedSingleNodeKafkaCluster CLUSTER = new EmbeddedSingleNodeKafkaCluster();
```
vvcephei: This final example still depends on the embedded Kafka _broker_, but now uses the mocked SR client.
All the microservice tests will follow this example.

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -51,18 +50,28 @@
...
 
+  // Create a mocked schema registry for our serdes to use
```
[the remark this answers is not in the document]
ableegoldman: Nice! My one comment is that it seems a little awkward to be directly accessing the private static CLIENT in order to register schemas: did you consider adding a #register method to the interface?

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -79,9 +88,8 @@ public void shouldRoundTripGenericAvroDataThroughKafka() throws Exception {
     streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "generic-avro-integration-test");
     streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
     streamsConfiguration.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.ByteArray().getClass().getName());
-    streamsConfiguration.put("mock.schema.registry.client", mockSchemaRegistryClient);
-    streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, MockGenericAvroSerde.class);
-    streamsConfiguration.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, "dummy config");
+    streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, GenericAvroSerde.class);
+    streamsConfiguration.put(AbstractKafkaAvroSerDeConfig.MOCK_SCHEMA_REGISTRY_CONFIG, TestSchemaRegistry.class);
```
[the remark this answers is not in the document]
cadonna: What happens if both, `MOCK_SCHEMA_REGISTRY_CONFIG` and `SCHEMA_REGISTRY_URL_CONFIG`, are set?

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -51,18 +50,28 @@
...
   @Test
   public void shouldRoundTripGenericAvroDataThroughKafka() throws Exception {
-    final MockSchemaRegistryClient mockSchemaRegistryClient = new MockSchemaRegistryClient();
 
     final Schema schema = new Schema.Parser().parse(
       getClass().getResourceAsStream("/avro/io/confluent/examples/streams/wikifeed.avsc")
     );
 
-    mockSchemaRegistryClient.register("inputTopic-value", schema);
+    TestSchemaRegistry.CLIENT.register("inputTopic-value", schema);
```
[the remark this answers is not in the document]
bbejeck: +1

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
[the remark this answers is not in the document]
vvcephei: Yeah, you could use the method to get it. `register` doesn't work, though, because the avro serdes are going to make their own instances of this class. That's why the field has to be static, so that all instances get a reference to the same client.

In another discussion I had another idea about how to do this, which I think you might like better. I'll update the proposal.

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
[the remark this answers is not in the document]
vvcephei: A config validation exception. They must be exclusive.

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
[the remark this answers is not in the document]
vvcephei: sounds good. I'm going to refactor this a bit. I'll keep this in mind.

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -51,18 +49,23 @@
  */
 public class GenericAvroIntegrationTest {
 
+  // A mocked schema registry for our serdes to use
+  private static final String SCHEMA_REGISTRY_SCOPE = GenericAvroIntegrationTest.class.getName();
```
vvcephei: Instead of requiring test authors to create their own scopes by creating their own static client references, we are now just letting them give each registry a named scope. They can use the same name everywhere (and therefore get the same registry everywhere), or give every test its own uniquely named scope (which I'm demonstrating).

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -51,18 +49,23 @@
...
   @Test
   public void shouldRoundTripGenericAvroDataThroughKafka() throws Exception {
-    final MockSchemaRegistryClient mockSchemaRegistryClient = new MockSchemaRegistryClient();
 
     final Schema schema = new Schema.Parser().parse(
       getClass().getResourceAsStream("/avro/io/confluent/examples/streams/wikifeed.avsc")
     );
 
-    mockSchemaRegistryClient.register("inputTopic-value", schema);
+    final SchemaRegistryClient schemaRegistryClient = MockSchemaRegistry.getClientForScope(SCHEMA_REGISTRY_SCOPE);
```
vvcephei: This is how you get a client for the given scope (you actually just get back the same `MockSchemaRegistryClient` every time you call this method with the same scope).

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -95,12 +97,12 @@ public void shouldRoundTripGenericAvroDataThroughKafka() throws Exception {
...
-    final Serde<GenericRecord> genericAvroSerde = new MockGenericAvroSerde(mockSchemaRegistryClient);
-    // Note how we must manually call `configure()` on this serde to configure the schema registry
-    // url.  This is different from the case of setting default serdes (see `streamsConfiguration`
+    final Serde<GenericRecord> genericAvroSerde = new GenericAvroSerde();
+    // Note how we must manually call `configure()` on this serde to configure the schema registry.
+    // This is different from the case of setting default serdes (see `streamsConfiguration`
     // above), which will be auto-configured based on the `StreamsConfiguration` instance.
     genericAvroSerde.configure(
-        Collections.singletonMap(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, "dummy config"),
+        Collections.singletonMap(AbstractKafkaAvroSerDeConfig.MOCK_SCHEMA_REGISTRY_CONFIG, SCHEMA_REGISTRY_SCOPE),
```
vvcephei: Now, instead of passing in your own "mock registry" subclass, you just pass in your desired scope name. This is actually a very similar concept to the schema registry url parameter, since the scope name uniquely references a single (mocked) Schema Registry instance.

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -128,11 +130,12 @@ public void shouldRoundTripGenericAvroDataThroughKafka() throws Exception {
         outputTopic,
         topologyTestDriver,
         new IntegrationTestUtils.NothingSerde<>(),
-        new KafkaAvroDeserializer(mockSchemaRegistryClient)
+        new KafkaAvroDeserializer(schemaRegistryClient)
       ).stream().map(kv -> (GenericRecord) kv.value).collect(Collectors.toList());
       assertThat(actualValues).isEqualTo(inputValues);
     } finally {
       topologyTestDriver.close();
+      MockSchemaRegistry.dropScope(SCHEMA_REGISTRY_SCOPE);
```
vvcephei: To not leak memory, you can drop the scope when you're done with it, so it can get garbage collected. But if you forget, it's not too bad, as it only pollutes the JVM heap while the tests are running.

src/test/java/io/confluent/examples/streams/SessionWindowsExampleTest.java
```diff
@@ -16,126 +16,118 @@
...
 
   @Before
   public void createStreams() {
-    streams =
-        SessionWindowsExample.createStreams(CLUSTER.bootstrapServers(),
-                                            CLUSTER.schemaRegistryUrl(),
-                                            TestUtils.tempDirectory().getPath());
-    streams.start();
+    // cleaning up beforehand just in case there's another test with the same scope.
+    MockSchemaRegistry.dropScope(SCHEMA_REGISTRY_SCOPE);
```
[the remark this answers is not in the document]
vvcephei: yeah, it's probably not needed at all in most circumstances... I'm having second thoughts about how weird this is as an example to set, although I do tend to write tests like this in practice, out of paranoia.

I think I'll just remove it from the "before"

vvcephei:
Depends on https://github.com/confluentinc/schema-registry/pull/1141 and https://github.com/confluentinc/common/pull/206

vvcephei:
retest this, please

vvcephei:
Hey @bbejeck @cadonna @ableegoldman and @LeilaYammine,

The upstream contribution was merged, after a slight design change. Now, we're re-using the same schema.registry.url config with a `mock://` protocol to define the scopes.

Do you mind giving this PR a final pass?

Thanks!
-John

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -95,12 +97,12 @@ public void shouldRoundTripGenericAvroDataThroughKafka() throws Exception {
...
-    final Serde<GenericRecord> genericAvroSerde = new MockGenericAvroSerde(mockSchemaRegistryClient);
-    // Note how we must manually call `configure()` on this serde to configure the schema registry
-    // url.  This is different from the case of setting default serdes (see `streamsConfiguration`
+    final Serde<GenericRecord> genericAvroSerde = new GenericAvroSerde();
+    // Note how we must manually call `configure()` on this serde to configure the schema registry.
+    // This is different from the case of setting default serdes (see `streamsConfiguration`
     // above), which will be auto-configured based on the `StreamsConfiguration` instance.
     genericAvroSerde.configure(
-        Collections.singletonMap(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, "dummy config"),
+        Collections.singletonMap(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, "mock://" + SCHEMA_REGISTRY_SCOPE),
```
cadonna: I like that more than a separate config parameter.

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
```diff
@@ -79,9 +82,8 @@ public void shouldRoundTripGenericAvroDataThroughKafka() throws Exception {
     streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "generic-avro-integration-test");
     streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
     streamsConfiguration.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.ByteArray().getClass().getName());
-    streamsConfiguration.put("mock.schema.registry.client", mockSchemaRegistryClient);
-    streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, MockGenericAvroSerde.class);
-    streamsConfiguration.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, "dummy config");
+    streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, GenericAvroSerde.class);
+    streamsConfiguration.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, "mock://" + SCHEMA_REGISTRY_SCOPE);
```
cadonna: I would put `"mock://" + SCHEMA_REGISTRY_SCOPE` into a String constant, because you also use it a couple of lines below.

src/test/java/io/confluent/examples/streams/interactivequeries/kafkamusic/KafkaMusicExampleTest.java
```diff
@@ -108,10 +116,10 @@ public static void createTopicsAndProduceDataToInputTopics() throws Exception {
     // Produce sample data to the input topic before the tests starts.
     final Properties props = new Properties();
     props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, CLUSTER.bootstrapServers());
-  
+
     final Map<String, String> serdeConfig = Collections.singletonMap(
-        AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, CLUSTER.schemaRegistryUrl());
-  
+        AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, "mock://" + SCHEMA_REGISTRY_SCOPE
```
cadonna: Same as above

src/test/java/io/confluent/examples/streams/GenericAvroIntegrationTest.java
vvcephei: sgtm

vvcephei:
Thanks for the review, @cadonna . I've addressed your feedback.

PR #262, opened 2019-06-23
fix: docker container kafka-music-data-generator not able to up due t…

habemusne:


habemusne:
[clabot:check]

andrewegel:
Causing CI to break due to the "unknown repository" - This needs to be closed to get around that. If this change is still required, please re-do it.

PR #270, opened 2019-07-25
Add new Aggregate and Reduce examples

miguno:
This is a subset of the original PR at https://github.com/confluentinc/kafka-streams-examples/pull/235 that contains only the new Aggregate + Reduce tests in Java and Scala, plus an updated IntegrationTestUtils (Java) and a new IntegrationTestScalaUtils.


miguno:
@vvcephei : This is the PR requested by you a while back.

https://github.com/confluentinc/kafka-streams-examples/pull/235#issuecomment-502128782

> e.g., if you just add the new examples, and your new IntegrationTestUtils convenience function, but don't migrate anything else, it will merge cleanly to future changes in older branches.

Let me know if this PR is better.

PR #273, opened 2019-07-26
Override default topic replication factor

### LINKED ISSUE #272: Unable to start kafka, since replication factor not set
With (at least) the 5.3.0 branch the single node kafka instance seems not to be configured properly, but expects the default of three nodes to be available, therefore the logs shows the error messages:

`ERROR [KafkaApi-1] Number of alive brokers '1' does not meet the required replication factor '3' for the offsets topic (configured via 'offsets.topic.replication.factor'). This error can be ignored if the cluster is starting up and not all brokers are up yet. (kafka.server.KafkaApis)`

nikos:
This allows the single-node Kafka cluster to start and fixes #272


ableegoldman:
Thanks for the fix! Do you mind retargeting this PR against 5.2.1-post so we can merge it forward from there up to 5.3.0?

Out of curiosity, were you running the music demo app?

nikos:
[clabot:check]

nikos:
> Thanks for the fix! Do you mind retargeting this PR against 5.2.1-post so we can merge it forward from there up to 5.3.0?

I tried to use the github feature to switch the base to `5.2.1-post` ... unfortunately it turned out to include more than 200 other commits with this simple change. So it is probably best to close this PR and open a clean new one against the `5.2.1-post` branch?

> Out of curiosity, were you running the music demo app?

Yep, exactly.

mjsax:
You would need to rebase the PR -- or maybe do a new one? Just redirecting to an older target branch won't work as you observed.

ableegoldman:
No worries, I went ahead and opened a new PR -- should be fixed on 5.3.0-post now. Thanks!

ableegoldman:
@nikos  FYI, If you're running the music demo app you may still be unable to view results outside the current container on the 5.3 branch. I'm still investigating this, but for the time being you can
a) just open a terminal in the music app container (docker exec -it {CONTAINER ID } /bin/bash)
b) modify the docker-compose and set the `kafka-music-application` containter's image to `confluentinc/kafka-streams-examples:5.0.0`

PR #278, opened 2019-08-03
Set KAFKA_MUSIC_APP_REST_HOST to 0.0.0.0

ableegoldman:
The music demo app has been "broken" for kafka-streams-examples images 5.1.0 and later, where the IQ results were not visible outside the container. 

This PR is targeted at 5.2.0-post since 5.1.0-post mistakenly uses the 5.0.0 image -- will fix in a follow up PR


ableegoldman:
The weird thing is that this only affects kafka-streams-examples versions 5.1.0+ (using  0.0.0.0 actually causes curl to hang for 5.0 and lower) -- not sure what changed to cause this (Dockerfile did switch from cp-base to cp-base-new between 5.0 and 5.1 but that seems about it?)

ableegoldman:
Merged with pint

PR #279, opened 2019-08-04
Fix empty response from server when running curl on music demo app

pantrif:
This allows curling the music demo app outside of the running container.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -130,7 +130,7 @@ services:
       STREAMS_BOOTSTRAP_SERVERS: kafka:29092
       STREAMS_SCHEMA_REGISTRY_HOST: schema-registry
       STREAMS_SCHEMA_REGISTRY_PORT: 8081
-      KAFKA_MUSIC_APP_REST_HOST: 0.0.0.0
+      KAFKA_MUSIC_APP_REST_HOST: kafka-music-application
       KAFKA_MUSIC_APP_REST_PORT: 7070
     extra_hosts:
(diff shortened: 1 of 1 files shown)


ableegoldman:
@pantrif Hi! You submitted this PR fairly soon after we merged (we think) a fix for this issue, where we changed `KAFKA_MUSIC_APP_REST_HOST` from `localhost` to `0.0.0.0`. I just want to confirm, were you hitting this issue before or also after we made the above change?

ie did you test the demo app with host `0.0.00` or was it still localhost when running into this?

ableegoldman:
Regardless of the above, we do ultimately want to set this to `kafka-music-application` as you've done. I believe this patch should be applied to 5.2.0 branch and later -- could you retarget the PR against 5.2.0-post so we can merge it downstream from there? You may need to rebase, or feel free to just open a new PR and ping us.

Thanks!

pantrif:
Hello @ableegoldman!! Indeed I submit this fix a few minutes after you merged. The localhost part was the problematic one. So I was hitting this issue before you make the above change.
However, I tried a few things of my own like setting the host to `127.0.0.1` or to `0.0.0.0` like you did and both seems to work just fine.
But I believe since you have already set up a host in the docker-compose file (`hostname: kafka-music-application`) I found better to provide that host (because 0.0.0.0 is too generic)

mjsax:
Closing this 5-year old PR. Version 5.x are EOL already.

We also migrated the content of this repo to https://github.com/confluentinc/tutorials in the mean time.

PR #281, opened 2019-08-05
Update compatibility matrix for post branch

ableegoldman:


ableegoldman:
@confluentinc/streams What even is the purpose of this matrix? All the information it contains could be conveyed in one line by saying Confluent versions numbers correspond to AK 3 major versions lower (eg CP 5.3.0 -> AK 2.3.0)

But instead we have this matrix we have to remember to update and format correctly, and confusingly it has some bugfix branches/versions listed but not others?

ableegoldman:
Someone from the tools team is helping to automatically update the version numbers correctly across kafka-streams-examples going forward, but this is the one nontrivial update when a branch is cut. I'd vote to just remove it, unless we feel it's actually serving some purpose or helping anyone

mjsax:
Atm it's three major version, but it might change at any point in time.

Also, the fact that it is 3 major version is know by people who use the platform already, but this information is especially important for new users.

mjsax:
Also, we don't need to update old version, but only need to add new ones. Hence, IMHO, the overhead is acceptable.

ableegoldman:
Fair enough. It would have been nice to close the manual updates gap for new versions but it's not like it's a prohibitive amount of work

ableegoldman:
merged with pint

PR #282, opened 2019-08-06
Fix empty response from server when running curl on music demo app

pantrif:
@ableegoldman I created a different PR retargeting against 5.1.3-post. Same with #279 with a different base branch.


ableegoldman:
Looks good, thanks for the patch! Will merge once we get a green build

PR #286, opened 2019-08-19
MINOR: Update timeout

bbejeck:
Although not a great practice, some of the restful integration tests could stand a bump in the timeout. I am doing this in `3.3.0-post` as we've seen the same error in `4.x` and `5.x` branches.


bbejeck:
> but do we really think it will stabilize the tests?

I think it will help some.  There were a few failures on the branches with lower timeouts.  This PR makes the timeouts consistent all the way up to master.

bbejeck:
merged with pint

PR #296, opened 2019-10-23
CustomStreamTableJoinIntegrationTest: Correctly read and validate test output

miguno:


src/test/java/io/confluent/examples/streams/CustomStreamTableJoinIntegrationTest.java
```diff
@@ -72,6 +72,10 @@
  *
  * Note: This example works with Java 8+ only.
  *
+ * TODO (and note for maintainers): This test can be migrated to use the TopologyTestDriver of Kafka Streams only once
+ * the driver is handling timestamps exactly like a "real" setup does. Until then, this test must use an embedded
```
bbejeck: What are the differences in how TTD handles timestamps?

src/test/java/io/confluent/examples/streams/CustomStreamTableJoinIntegrationTest.java
```diff
@@ -280,8 +284,10 @@ public void shouldTriggerStreamTableJoinFromTable() throws Exception {
       //
       // Step 3: Verify the application's output data.
       //
-      final List<KeyValue<String, Long>> actualRecords = readOutputDataFromJoinedStream(expectedOutputRecords.size());
+      final List<KeyValue<String, Pair<Double, Long>>> actualRecords =
+          readOutputDataFromJoinedStream(expectedOutputRecords.size());
       assertThat(actualRecords).isEqualTo(expectedOutputRecords);
+      assertThat(actualRecords).containsExactlyElementsOf(expectedOutputRecords);
```
bbejeck: nit: with this check do we need the `isEqualTo` check above still?

miguno: Not needed, removed in follow-up commit.

src/test/java/io/confluent/examples/streams/CustomStreamTableJoinIntegrationTest.java
miguno: IIRC the difference is that TTD processes each event synchronously in exactly the order you pass them, whereas KStreams buffers both inputs (of a join) and selects records based on the lowest timestamp. @vvcephei knows more.

vvcephei: This is correct. The comment implies this is a flaw and maybe that there are plans to change it, but AFAIK, there are no plans to change it, and I'd tentatively say it's a feature, not a bug.

If TTD behaved _exactly_ like a real setup in this regard, it would be just as difficult to write reliable tests as it is against a real setup.

Specifically in regard to whether this test can be migrated or not, would it not work to just pipe the inputs in the same order as their timestamps?

PR #298, opened 2019-10-28
MINOR: add missing deprecation annotation

mjsax:


src/test/java/io/confluent/examples/streams/window/DailyTimeWindows.java
```diff
@@ -52,6 +52,7 @@ public DailyTimeWindows(final ZoneId zoneId, final int startHour, final Duration
         return windows;
     }
 
+    @Deprecated
```
cpettitt-confluent: Could we rework this to use non-deprecated functionality, especially since this is a user example? At the very least we should tell the user what they should use instead, but getting them started with the right approach would be much better!

vvcephei: +1. In fact, the comment in the method gives the way to do it, use `Materialize.withRetention`.

mjsax: Well, `DailyWindowes extends Windows` and thus it must implement the method... Not sure what we could do about it?

cpettitt-confluent: I was proposing to rework the example in such a way that it uses functionality that we support. I'm assuming that means not extending `Windows`, but using `Materialize.withRetention`?

vvcephei: no, the example requires us to extend Windows, but I don't think we have to override this method.

cpettitt-confluent: Thanks @vvcephei, that makes sense! And my being the naive user goes to show why fixing this would be a good idea :).

src/test/java/io/confluent/examples/streams/window/DailyTimeWindows.java
```diff
@@ -52,16 +52,6 @@ public DailyTimeWindows(final ZoneId zoneId, final int startHour, final Duration
         return windows;
     }
 
-    @Override
-    public long maintainMs() {
```
mjsax: @jeanlouisboudart @csarrazi 

This method should not be overwritten. Instead, the store retention time should be increase by the user of `DailyTimeWindows`

mjsax:
@vvcephei @cpettitt-confluent Updated this PR.

PR #299, opened 2019-10-28
MINOR: fix deprecation warnings

mjsax:


src/main/java/io/confluent/examples/streams/microservices/FraudService.java
```diff
@@ -75,6 +75,7 @@ public void start(final String bootstrapServers, final String stateDir) {
     log.info("Started Service " + getClass().getSimpleName());
   }
 
+  @SuppressWarnings("unchecked")
```
mjsax: Suppress warning about untypes `branch()` operator.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -94,13 +96,13 @@
   public static <K, V> List<KeyValue<K, V>> readKeyValues(final String topic, final Properties consumerConfig, final int maxMessages) {
     final KafkaConsumer<K, V> consumer = new KafkaConsumer<>(consumerConfig);
     consumer.subscribe(Collections.singletonList(topic));
-    final int pollIntervalMs = 100;
-    final int maxTotalPollTimeMs = 2000;
+    final Duration pollInterval = Duration.ofMillis(100L);
+    final int maxTotalPollTimeMs = 10000;
```
mjsax: The new `Consumer.poll(Duration)` has different semantics that the deprecated `poll(long)` -- needed to increase the timeout to make tests passing locally.

src/test/java/io/confluent/examples/streams/microservices/util/MicroserviceTestUtils.java
```diff
@@ -133,7 +133,7 @@ public static void stopTailers() {
   }
 
   public static void tailAllTopicsToConsole(final String bootstrapServers) {
-    for (final Topic t : Topics.ALL.values()) {
+    for (final Topic<?, ?> t : Topics.ALL.values()) {
```
mjsax: Avoid warning about raw-type.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -94,13 +96,13 @@
...
     int totalPollTimeMs = 0;
     final List<KeyValue<K, V>> consumedValues = new ArrayList<>();
     while (totalPollTimeMs < maxTotalPollTimeMs && continueConsuming(consumedValues.size(), maxMessages)) {
-      totalPollTimeMs += pollIntervalMs;
-      final ConsumerRecords<K, V> records = consumer.poll(pollIntervalMs);
+      totalPollTimeMs += pollInterval.toMillis();
```
vvcephei: would it be better to just increase this by the actual time spent in `poll`?

src/main/java/io/confluent/examples/streams/microservices/FraudService.java
vvcephei: Can you elaborate? This sounds like we designed a flaw into the API, such that anyone using `branch` would have to suppress this compiler warning.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: I think if we adjust the way we increment `totalPollTimeMs`, we can leave this maximum alone.

src/main/java/io/confluent/examples/streams/microservices/FraudService.java
cpettitt-confluent: Could we do https://docs.oracle.com/javase/7/docs/api/java/lang/SafeVarargs.html on the `branch` interface, since this would be required by every caller? If not, it would still be cleaner to move this line to the call site to minimize the scope of warning suppression.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
mjsax: Can you elaborate?

src/main/java/io/confluent/examples/streams/microservices/FraudService.java
cpettitt-confluent: Sounds good. I would prefer to fix this properly too. In the meantime, I would suggest moving the annotation down to the statement that has the problem versus doing it at method scope.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
mjsax: That does not work. I dug into this a little bit, and the old `poll(long)` takes about 3000ms the first time it's called. Hence, the timeout of 2000ms is too low for the new methods that block about 100ms the first time. Getting the metadata from the brokers is just more overhead we need to factor in.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
mjsax: We can do this -- even if it does not change much, because we stay about 100ms in `poll(Duration)` per call anyway.

mjsax:
Updated this. @vvcephei @cpettitt-confluent

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
```diff
@@ -110,6 +114,7 @@
   }
 
   private static boolean continueConsuming(final int messagesConsumed, final int maxMessages) {
+    System.err.println("mjsax: " + messagesConsumed + " -- " + maxMessages);
```
cpettitt-confluent: Oops :)

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: I see. Thanks for the explanation!

src/main/java/io/confluent/examples/streams/microservices/FraudService.java
vvcephei: A statement-level suppression may satisfy IDEA but not the compiler, the only way to know is to try.

Regardless, I'm fine with adding the suppression. Thanks for the explanation.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
vvcephei: I'll leave it to you.

Since the timeout for the test is defined in terms of the time spent, it might make more sense to use the time spent. Otherwise it's effectively just counting the _number_ of polls, but in a roundabout way, where the max number of polls is expressed as `maxTotalPollTimeMs / pollInterval.toMillis` and each time, we count up by incrementing `pollInterval.toMillis` amount.

It's a little janky as-is, but not a major concern for me.

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
mjsax: Arg. I though I removed all debug stuff...

src/test/java/io/confluent/examples/streams/IntegrationTestUtils.java
mjsax: Yes. Updated as request. Does not make a big difference, but as you said, it's more precise to actually measure it, instead of counting poll-calls.

src/main/java/io/confluent/examples/streams/microservices/FraudService.java
mjsax: `mvn clean compile test-compile` did not give a warning any longer, even with the statement-level suppression -- the compiler seems to be happy :)

mjsax:
Updated this again. Hope it's ok now. @vvcephei @cpettitt-confluent

PR #300, opened 2019-10-29
MINOR: bump Scala version

mjsax:


pom.xml
```diff
@@ -56,7 +56,7 @@
         <java.version>1.8</java.version>
         <licenses.version>5.4.0-SNAPSHOT</licenses.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <scala.version>${kafka.scala.version}.7</scala.version>
+        <scala.version>${kafka.scala.version}.8</scala.version>
```
mjsax: Seems some dependency changed it Scala version from `2.12.7` to `2.12.8` which are not binary compatible. Bumping version to fix `master`.

PR #302, opened 2019-10-30
Fix Slack integration

mjsax:
Follow up to #290

Without a slack channel specified, it falls back to an incorrect default channel.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,4 +11,5 @@ docker_oraclejdk8 {
     // nodeLabel = 'docker-oraclejdk8-compose'
     upstreamProjects = 'confluentinc/common'
     withPush = true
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -11,4 +11,5 @@ docker_oraclejdk8 {
     // nodeLabel = 'docker-oraclejdk8-compose'
     upstreamProjects = 'confluentinc/common'
     withPush = true
+    slackChannel = 'streams-alerts'
```
mjsax: We could also disable it completely if we set it to `''` -- thoughts?

mjsax:
Closing this because this should be "fixed" in a different repository to not have a default notification channel.

PR #306, opened 2019-12-09
ST-2560: Adding support for building debian and rhel images

elismaga:
Since this image is built from artifacts and not packages we don't need to create separate docker files for each os type we want to build, which is the pattern we use with the other images that are built from packages. This change will build the same docker file twice but each time using a different docker upstream tag so that it uses a different base image each time. 

I have tested the debian images locally and it worked with cp-demo. Testing of the rhel images with cp-demo is still on going.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,4 +7,6 @@ dockerfile {
     upstreamProjects = 'confluentinc/rest-utils'
     nodeLabel = 'docker-oraclejdk8-compose-swarm'
     cron = ''
+    cpImages = true
+    osTypes = ['deb9', 'rhel8']
 }
--- a/pom.xml
+++ b/pom.xml
@@ -58,6 +58,10 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <scala.version>${kafka.scala.version}.8</scala.version>
         <scalatest.version>3.0.5</scalatest.version>
+        <docker.os_type>deb9</docker.os_type>
+        <!-- Need to explicitly set this otherwise it will be overridden from common-docker pom. -->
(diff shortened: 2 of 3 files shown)


Jenkinsfile
```diff
@@ -7,4 +7,6 @@ dockerfile {
     upstreamProjects = 'confluentinc/rest-utils'
     nodeLabel = 'docker-oraclejdk8-compose-swarm'
     cron = ''
+    cpImages = true
+    osTypes = ['deb9', 'rhel8']
```
mohnishbasha: are we planning to support deb9 for this change?

pom.xml
```diff
@@ -58,6 +58,10 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <scala.version>${kafka.scala.version}.8</scala.version>
         <scalatest.version>3.0.5</scalatest.version>
+        <docker.os_type>deb9</docker.os_type>
```
mohnishbasha: do we want to build for deb9 as default?

Jenkinsfile
elismaga: @mohnishbasha This image has always used a Debian9 base image. cp-base-new has always been debian 9, which is what these images were always based on.

pom.xml
elismaga: The default for this image should be deb9 because that is what this image has always used. this image was always based on cp-base-new which is a Debian 9 image.

mohnishbasha:
now that we have rhel, do we need to update anything on readme?

elismaga:
> now that we have rhel, do we need to update instructions on readme as well?

yes, there is a task on the one pager / plan to update the readme and the wiki docs. that will happen today or tomorrow

PR #310, opened 2019-12-31
Remove SNAPSHOT suffix from versions in comments

cadonna:
The removed SNAPSHOT suffixes were not automatically removed by
Jenkins.

For an example of the SNAPSHOT suffixes removed by Jenkins
see
https://github.com/confluentinc/kafka-streams-examples/commit/fa1a5e1a159a7777912b8e0ae10dcf871a06d183


ableegoldman:
I wonder why the Jenkins job missed just these few cases? Maybe we should file a ticket with the Tools team?

cadonna:
I am also wondering about that. I have already filed a ticket https://confluentinc.atlassian.net/browse/ST-2836.

PR #311, opened 2019-12-31
Update version compatibility matrix

cadonna:
- remove SNAPSHOT suffixes
- add versions 5.3.1 and 5.3.2
- remove wrong foot note

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,7 +379,9 @@ $ mvn package # Packages the application examples into a standalone jar
 
 | Branch (this repo)                      | Confluent Platform | Apache Kafka      |
 | ----------------------------------------|--------------------|-------------------|
-| [5.4.x](../../../tree/5.4.x/)\*         | 5.4.0-SNAPSHOT     | 2.4.0-SNAPSHOT    |
+| [5.4.0-post](../../../tree/5.4.0-post/) | 5.4.0              | 2.4.0             |
+| [5.3.2-post](../../../tree/5.3.0-post/) | 5.3.2              | 2.3.1             |
+| [5.3.1-post](../../../tree/5.3.0-post/) | 5.3.1              | 2.3.0             |
 | [5.3.0-post](../../../tree/5.3.0-post/) | 5.3.0              | 2.3.0             |
 | [5.2.2-post](../../../tree/5.2.2-post/) | 5.2.2              | 2.2.1             |
 | [5.2.1-post](../../../tree/5.2.1-post/) | 5.2.1              | 2.2.1             |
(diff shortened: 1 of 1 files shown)


mjsax:
This is not streams related -- I can't really review this.

README.md
```diff
@@ -379,7 +379,9 @@ $ mvn package # Packages the application examples into a standalone jar
 
 | Branch (this repo)                      | Confluent Platform | Apache Kafka      |
 | ----------------------------------------|--------------------|-------------------|
-| [5.4.x](../../../tree/5.4.x/)\*         | 5.4.0-SNAPSHOT     | 2.4.0-SNAPSHOT    |
+| [5.4.0-post](../../../tree/5.4.0-post/) | 5.4.0              | 2.4.0             |
```
mjsax: There are many other versions missing... It is actually worth to list all bug fix releases?

Also 5.4.0-post does not exist yet -- should we do this change only after the release?

README.md
```diff
@@ -389,8 +391,6 @@ $ mvn package # Packages the application examples into a standalone jar
 | [4.0.0-post](../../../tree/4.4.0-post/) | 4.0.0              | 1.0.0             |
 | [3.3.0-post](../../../tree/3.3.0-post/) | 3.3.0              | 0.11.0            |
 
-\*You must manually build the `2.3` version of Apache Kafka and the `5.3.x` version of Confluent Platform.  See instructions above.
-
```
mjsax: Why remove this? Should have been updated to `2.4` I guess (and should be updated to `2.5` now?)

cadonna: I removed the footnote because with this change there is no SNAPSHOT version anymore in the matrix, i.e. no footnote is needed anymore. It could be confusing having a footnote w/o a reference to it.

What we should do in any case is making sure that the footnote is correct on master and on the 5.5.x branch that will be cut from it before the next release, because on those branches we will have again a SNAPSHOT version.

mjsax: Well, we would also have a `5.4.1-SNAPSHOT` on `5.4.x` branch -- but I guess you imply that we should not refer to snapshot versions for those, what sounds reasonable to me.

For this case, we would only have a snapshot reference on `master` because `5.5.x` will only be cut after code freeze.

README.md
mjsax: Sounds reasonable. Thanks!

About bug-fix releases and what AK version: this is all documented in CP docs, hence, I am wondering if we should duplicate it here? Maybe better add a link to CP docs? It's already quite hard to keep the information up to date in one place.

mjsax: Should we also update older branches accordingly? (Maybe going back to 5.0.x?)

PR #312, opened 2020-01-10
MINOR: Change Avro dependency required for schema registry updates

bbejeck:
Schema registry added pluggable schemas in master, `Avro` dependency needed to get upgraded to 1.9.1.  This PR is needed to fix the kafka-streams-examples master build failures.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,7 +49,7 @@
     -->
     <properties>
         <algebird.version>0.13.4</algebird.version>
-        <avro.version>1.8.2</avro.version>
+        <avro.version>1.9.1</avro.version>
         <chill.version>0.9.2</chill.version>
         <docker.skip-build>false</docker.skip-build>
         <docker.skip-test>false</docker.skip-test>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -158,6 +158,11 @@
             <artifactId>jackson-annotations</artifactId>
             <version>${jackson.version}</version>
         </dependency>
+        <dependency>
```
bbejeck: This needed to get added, as before it seems that `ObjectMapper` was pulled in as a transitive dependency of the old Avro version.  This dependency should have been added in the first place.

src/main/java/io/confluent/examples/streams/JsonToAvroExample.java
```diff
@@ -15,20 +15,21 @@
  */
 package io.confluent.examples.streams;
 
+import com.fasterxml.jackson.databind.JsonNode;
```
bbejeck: same reason for changes here

src/main/java/io/confluent/examples/streams/JsonToAvroExampleDriver.java
```diff
@@ -15,17 +15,11 @@
  */
 package io.confluent.examples.streams;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
```
bbejeck: ditto

bbejeck:
Merged #312 into master

PR #329, opened 2020-04-21
Replace `./gradlew installAll` with `./gradlewAll install and remove invocation of `gradle`

ijuma:
See https://github.com/apache/kafka/pull/7677 for more details with regards to
the former and https://github.com/apache/kafka/pull/6031 with regards to the latter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,11 +185,8 @@ repository may have different Kafka requirements.
 > $ cd kafka
 > $ git checkout trunk
 >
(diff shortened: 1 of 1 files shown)


mjsax:
@ijuma The AK PR was against `trunk` (and I think not cherry-picked to older branches) -- hence, should this PR be against `master` only instead of `5.4.1` ?

ijuma:
Yes, that's what I intended.

PR #330, opened 2020-04-23
KSTREAMS-4026: Fix type parameter bounds of ProcessorContext in Scala

cadonna:
The AK PR https://github.com/apache/kafka/pull/8414 introduced type parameters
for the ProcessorContext. Since Scala requires generic arguments for any
type that takes generic parameters, the Scala code in the examples that uses
the ProcessorContext did not compile anymore.

### DIFF
--- a/src/main/scala/io/confluent/examples/streams/algebird/CMSStore.scala
+++ b/src/main/scala/io/confluent/examples/streams/algebird/CMSStore.scala
@@ -185,7 +185,7 @@ class CMSStore[T: CMSHasher](override val name: String,
   /**
     * Initializes this store, including restoring the store's state from its changelog.
     */
-  override def init(context: ProcessorContext, root: StateStore) {
+  override def init(context: ProcessorContext[_, _], root: StateStore) {
     val serdes = new StateSerdes[Integer, TopCMS[T]](
       name,
       Serdes.Integer(),
--- a/src/main/scala/io/confluent/examples/streams/algebird/CMSStoreChangeLogger.scala
+++ b/src/main/scala/io/confluent/examples/streams/algebird/CMSStoreChangeLogger.scala
@@ -29,14 +29,14 @@ import org.apache.kafka.streams.state.StateSerdes
   * with the [[org.apache.kafka.common.utils.Bytes]] class.
   */
 class CMSStoreChangeLogger[K, V](val storeName: String,
-                                 val context: ProcessorContext,
+                                 val context: ProcessorContext[_, _],
                                  val partition: Int,
                                  val serialization: StateSerdes[K, V]) {
 
   private val topic = ProcessorStateManager.storeChangelogTopic(context.applicationId, storeName)
   private val collector = context.asInstanceOf[RecordCollector.Supplier].recordCollector
 
-  def this(storeName: String, context: ProcessorContext, serialization: StateSerdes[K, V]) {
+  def this(storeName: String, context: ProcessorContext[_, _], serialization: StateSerdes[K, V]) {
     this(storeName, context, context.taskId.partition, serialization)
   }
 
--- a/src/main/scala/io/confluent/examples/streams/algebird/ProbabilisticCounter.scala
+++ b/src/main/scala/io/confluent/examples/streams/algebird/ProbabilisticCounter.scala
@@ -11,9 +11,9 @@ class ProbabilisticCounter(val cmsStoreName: String)
   extends Transformer[String, String, KeyValue[String, Long]] {
 
   private var cmsState: CMSStore[String] = _
-  private var processorContext: ProcessorContext = _
+  private var processorContext: ProcessorContext[Object, Object] = _
 
-  override def init(processorContext: ProcessorContext): Unit = {
+  override def init(processorContext: ProcessorContext[Object, Object]): Unit = {
     this.processorContext = processorContext
     cmsState = this.processorContext.getStateStore(cmsStoreName).asInstanceOf[CMSStore[String]]
   }


mjsax:
Does this expose an unexpected backward incompatibility?  \cc @vvcephei

cadonna:
According to https://github.com/apache/kafka/pull/8414#issuecomment-608644050 it is not unexpected.

mjsax:
Thanks for the reference. Not sure how to proceed...

vvcephei:
Thanks, all. I'm not sure how to proceed either. Let's resume the discussion here:

>Hmm, that last commit illustrates a problem with this approach. Adding generics to a previously raw type is only a compiler warning for Java users, but it's a compiler error for Scala users, because Scala insists on having the generic arguments for any type that takes generic parameters.
>
>I think the way to fix this is to propose to add a new ProcessorContext as well as a new Processor interface. Or we can accept this level of source code incompatibility in the Scala API. For context, many changes we've made in minor versions break Scala source compatibility, since Scala generally makes it harder for libraries to implement non-breaking deprecation paths.

Clearly, the choice I made before is that it's ok to break in this way, but maybe now it seems "a little too real". Should we instead take the alternative approach I proposed?

mjsax:
I don't know Scala very well. But if you can fit it in the Scala API, it seems reasonable to me?

vvcephei:
We can't fit it into the Scala API because these are types from the Java API. My alternative proposal was to create a brand new _Java_ ProcessorContext, similar to how we are going to create a brand-new Processor. Basically, we leave ProessorContext untouched and replace it with a generically-typed one.

mjsax:
Not sure if I can follow: are you saying we would need to revert the AK PR that added generics?

vvcephei:
Yes, we'd revert that change and add a new ProcessorContext with generics, and then deprecate the old one. The thing I don't like about that is that the scope of deprecating ProcessorContext is much larger than the scope of deprecating the Processor interface, since ProcessorContext is exposed even to DSL users in a number of places, such as Transformers.

I guess a middle ground is to introduce the new ProcessorContext without deprecating the old one, to limit the splash damage, while still supporting both backward compatibility and the change we want to make.

vvcephei:
Hey, @cadonna @mjsax ,

I'd like to propose this approach instead:

It seems like supporting backward compatibility and supporting cross-language usage are both feasible, but supporting backward compatibility WITH cross-language usage is potentially too much to ask.

I'd be more inclined just to hand-wave this particular breakage if we could say with confidence that it won't happen again. But I'm not sure we can say that. What would help is to create a "pure" Scala API. In other words, we would not leak any objects from the Java API through to the Scala API, so we'd have a Scala ProcessorContext, Scala Transformer, etc., etc. This would let us explicitly manage the APIs' backward-compatibility properties independently.

For example, in this case, we would already have had a Scala ProcessorContext (with no generics) that adapts the Java one. We would have had to update the Scala ProcessorContext _implementation_ to cope with the new generics, but wouldn't add the generics to the Scala ProcessorContext _interface_.

Shoring up the Java interfaces that already leak through to the Scala API would itself be a breaking change, although we would be able to transition most of the usages via deprecation. However, even though we'd still break ProcessorContext usages under this plan, it comes with a promise that we stand a better chance of avoiding such breakage in the future. Essentially, we'd be admitting we made a mistake in the API design, and that we have to break compatibility to some extent to fix the mistake.

Specifically, if you feel ok with this plan, I'd make a 2.6 blocker to do what I said above, and then we would proceed to merge this PR.

cadonna:
@vvcephei IMO it would be indeed more robust to have a dedicated Scala API that does not leak the Java API. The only concern that I have is that if we now merge this PR (and do not revert the other), we must complete the Scala API for the 2.6 release. Are we confident to get that done? Or are you saying we just accept this specific breakage in any case?

vvcephei:
Thanks for the reply, @cadonna .

Yes, just to be clear, this particular usage would break in any case. The “full Scala API” plan is just to soften the blow. I’d be willing to commit to it for 2.6 if that lets us make progress here.

mjsax:
Not sure how much work this would be? Maybe @apurvam has some thoughts on this, too, with regard to planning?

guozhangwang:
Just to clarify the context here, in https://cwiki.apache.org/confluence/display/KAFKA/KIP-478+-+Strongly+typed+Processor+API the only "in-place change" we proposed is `ProcessorContext`, while for all others we are introducing as either new classes or new APIs. So does this mean the only issue we are having is for `ProcessorContext`, and with this "full Scala Api" work, the only fix would be for ProcessorContext while others are actually okay to still get the (new, strong-typed) Java object leaked into it. Is that right?

vvcephei:
Yes, that's correct, @guozhangwang . The other thing I can do is investigate how much of an impact it would be to back out this change and instead introduce a new (typed) ProcessorContext.

vvcephei:
For some more context, I just talked with a community user of the scala API, and he indicated that he would be satisfied with the breakage, since it improves the type safety of the system. Adding a "shielding" layer to prevent Java types from leaking through the Scala interface also struck him as a really good idea.

On the other hand, he said that he only has a handful of references to ProcessorContext, so if he were a heavy user of PAPI, he might feel differently.

Of course, it's just one data point.

guozhangwang:
If the extra amount of work of maintaining this shielding in the Scala interface is not huge, I'm +1 for it :) Just want to make sure we are not unconsciously committing to a huge maintenance overhead moving forward.

vvcephei:
I think it won't be immediately necessary now that I'm backtracking on the ProcessorContext change, but the shielding would still do more good than harm in the long run. Specifically, maintaining it would be some amount of work, but the nature of the work should be pretty straightforward compared to tracking down and repairing _unintentional_ breaking API changes, like the one that this PR is for.

But we can defer that work for a later day, or even implement it in stages.

vvcephei:
Hey @cadonna , I've just reverted the upstream change, so this PR should no longer be necessary. Sorry for the trouble.

cf https://github.com/apache/kafka/pull/8595

PR #342, opened 2020-06-04
HOTFIX: revert #340 and disable spotbugs

vvcephei:
Since the upgrade to 2.13, spotbugs is choking on Scala's generated
bytecode. See #340 for the errors. 

The prior fix #340 unintentionally broke the examples by changing
the applications from `object` to `class`, which makes them no longer
runnable. I.e., this change converted the main method to a non-static
method, so there wouldn't be any way to actually run the examples
on the console.

Since this repo doesn't contain any production code (just examples),
we can just disable spotbugs. We still have compiler warnings as
errors and checkstyle.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -470,6 +470,14 @@
             </plugin>
 
             <plugin>
+                <groupId>com.github.spotbugs</groupId>
+                <artifactId>spotbugs-maven-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
+
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
                 <executions>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -469,6 +469,14 @@
                 </configuration>
             </plugin>
 
+            <plugin>
+                <groupId>com.github.spotbugs</groupId>
+                <artifactId>spotbugs-maven-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
```
mjsax, line 478: Do we really want do disable it completely? Can't we add exceptions for just the two files in question?

vvcephei: Sure, we can add exceptions instead, but I'm challenging the value of spotbugs for this example project:

> Since this repo doesn't contain any production code (just examples),
we can just disable spotbugs. We still have compiler warnings as
errors and checkstyle.

vvcephei: Note that we never _decided_ to add spotbugs to begin with, we just inherited it transitively through the parent poms.

mjsax: Well, writing good example code seems valuable, too :)

Don't have a strong opinion, but if it's company policy to use spotbugs, we might just align to it? \cc @guozhangwang 

Just sharing my opinion. I am also ok to completely disable it if we think that's best.

vvcephei: Thanks, @mjsax . There haven't been any further comments, so I'll just go ahead with this. We can always refine it later, but I'd rather not risk forgetting about this PR and releasing broken examples.

PR #350, opened 2020-07-24
DEVX-1867: microservices: added exception hdlr for PostOrdersAndPayments

rspurgeon:
This allows the service to continue to try and post orders while the OrdersService is unavailable
or not yet ready.

### DIFF
--- a/src/main/java/io/confluent/examples/streams/microservices/PostOrdersAndPayments.java
+++ b/src/main/java/io/confluent/examples/streams/microservices/PostOrdersAndPayments.java
@@ -138,29 +138,33 @@ public class PostOrdersAndPayments {
 
             // POST order to OrdersService
             System.out.printf("Posting order to: %s   .... ", path.urlPost());
-            final Response response = client.target(path.urlPost())
-                                            .request(APPLICATION_JSON_TYPE)
-                                            .post(Entity.json(inputOrder));
-            System.out.printf("Response: %s %n", response.getStatus());
+            try {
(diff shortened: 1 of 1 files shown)


src/main/java/io/confluent/examples/streams/microservices/PostOrdersAndPayments.java
```diff
@@ -138,29 +138,33 @@ public static void main(final String[] args) throws Exception {
...
+                        .request(APPLICATION_JSON_TYPE)
+                        .get(newBean());
+                if (!inputOrder.equals(returnedOrder)) {
+                    System.out.printf("Posted order %d does not equal returned order: %s%n", i, returnedOrder.toString());
+                } else {
+                    System.out.printf("Posted order %d equals returned order: %s%n", i, returnedOrder.toString());
+                }
+
+                // Send payment
+                final Payment payment = new Payment("Payment:1234", id(i), "CZK", 1000.00d);
```
mikebin, line 160: Is it ok to send the order above and fail sending the payment? Or should sending the payment have it's own try/catch block?

rspurgeon: Thanks Mike, I took the fastest route I could come up with.  The thing I was trying to solve was the order service http service not yet available, which is the common issue for the demos we use this in.  That's the first thing in this block so my assumption was it would fail there and then retry the whole block again.

PR #360, opened 2020-09-12
update the consumer property

ganesh-sai:
While trying to view the count of each message unable to view the numbers since kafka-console-consumer is not able to understand the property value.deserializer


ganesh-sai:
[clabot:check]

PR #372, opened 2021-01-21
Add an example using addGlobalStore and transform to join topics' data

adrien-ben:
Adds an example mimicking the `GlobalKTablesExample` but using `addGlobalStore` and `transformValues` to perform the "join".

In the `GlobalStoresExampleTest` I use the `GlobalKTablesExampleDriver` should I duplicate it to create a `GlobalStoresExampleDriver` ?


adrien-ben:
[clabot:check]

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
```diff
@@ -0,0 +1,246 @@
...
+    static final String CUSTOMER_STORE = "customer-store";
+    static final String PRODUCT_STORE = "product-store";
+    static final String ENRICHED_ORDER_TOPIC = "enriched-order";
+
+    public static void main(final String[] args) {
+        final String bootstrapServers = args.length > 0 ? args[0] : "localhost:9092";
+        final String schemaRegistryUrl = args.length > 1 ? args[1] : "http://localhost:8081";
+        final KafkaStreams
+                streams =
+                createStreams(bootstrapServers, schemaRegistryUrl, "/tmp/kafka-streams-global-tables");
```
spena, line 108: Shouldn't be `/tmp/kafka-streams-global-store`?

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
```diff
@@ -0,0 +1,246 @@
...
+            store = (KeyValueStore<K, V>) processorContext.getStateStore(storeName);
+        }
+
+        @Override
+        public void process(final K key, final V value) {
+            if (value != null) {
+                store.put(key, value);
+            } else {
+                store.delete(key);
+            }
```
spena, line 237: Would a single-line `store.put(key, value)` achieve the same thing? If `value` is null, then the `store.put(key, null)` will delete the corresponding key, won't it?

spena (commented):
Code looks good. I see a lot of duplicated effort, though. I'm not sure if having a generic or abstract class would help. This is a test case that may require different parameters in the future, so I'm not sure. What do you think?

src/test/java/io/confluent/examples/streams/GlobalStoresExampleTest.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @After
+    public void close() {
+        streamInstanceOne.close();
+        streamInstanceTwo.close();
+    }
+
+    @Test
+    public void shouldDemonstrateGlobalKTableJoins() throws Exception {
```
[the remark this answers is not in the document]
adrien-ben: Done

src/test/java/io/confluent/examples/streams/GlobalStoresExampleTest.java
```diff
@@ -0,0 +1,142 @@
...
+                IntegrationTestUtils.waitUntilMinValuesRecordsReceived(consumerProps, ENRICHED_ORDER_TOPIC, 50, 60000);
+
+        // verify that all the data comes from the generated set
+        for (final EnrichedOrder enrichedOrder : enrichedOrders) {
+            assertThat(customers, hasItem(enrichedOrder.getCustomer()));
+            assertThat(products, hasItem(enrichedOrder.getProduct()));
+            assertThat(orders, hasItem(enrichedOrder.getOrder()));
+        }
+
+        // demonstrate that global table data is available on all instances
```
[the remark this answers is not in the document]
adrien-ben: Done

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
adrien-ben: Absolutely. Changed.

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
adrien-ben: Yes, don't know how I did not see this... thank you

adrien-ben:
About duplication, I started by copy pasting the GlobalTables example and test classes and just replaced the KTables/join code by GlobalStores/transform. I thought about factoring some of the code but preferred to have the two examples separated to avoid the struggle of splitting them back if needed later. 

For the test classes maybe I could add a helper class contains the code that pushes orders/products/consumers and the assertions. What do you think ?

spena (approved):
Agree on the duplication issue. If we need to split this test later (change a config, or something else), then we'll end up doing more work. Well, the test looks good to me. I don't think we'd need to add a utils class.

@mjsax What do you think?

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
```diff
@@ -0,0 +1,242 @@
...
+import org.apache.kafka.streams.state.KeyValueStore;
+import org.apache.kafka.streams.state.Stores;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * Demonstrates how to perform "joins" between  KStreams and GlobalStore, i.e. joins that
+ * don't require re-partitioning of the input streams.
```
mjsax, line 44: Might be worth to add a sentence to point to the KStream-GlobalKTable example and explain that both effectively do the same thing? Also add a sentence to the exiting example to point to this new example?

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
```diff
@@ -0,0 +1,242 @@
...
+ * 3) Start this example application either in your IDE or on the command line.
+ * <p>
+ * If via the command line please refer to <a href='https://github.com/confluentinc/kafka-streams-examples#packaging-and-running'>Packaging</a>.
+ * Once packaged you can then run:
+ * <pre>
+ * {@code
+ * $ java -cp target/kafka-streams-examples-6.0.1-standalone.jar io.confluent.examples.streams.GlobalStoresExample
+ * }
+ * </pre>
+ * 4) Write some input data to the source topics (e.g. via {@link GlobalKTablesExampleDriver}). The
```
mjsax, line 79: Should we point out that we reuse this diver for two examples? Otherwise, the name might be confusing? Or rename the driver class?

mjsax (commented):
Thanks for the PR @adrien-ben!

About test code duplication: seems reasonable to me to avoid the duplicated code, as we literally test the same input data and expect the same result.

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
```diff
@@ -0,0 +1,242 @@
...
+        private KeyValueStore<K, V> store;
+
+        @Override
+        public void init(final ProcessorContext processorContext) {
+            store = (KeyValueStore<K, V>) processorContext.getStateStore(storeName);
+        }
+
+        @Override
+        public void process(final K key, final V value) {
+            store.put(key, value);
```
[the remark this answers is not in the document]
adrien-ben: Done

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
adrien-ben: Added a small paragraph in both example referencing the other and how they differ.

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
adrien-ben: Updated the doc and renamed the class `GlobalKTablesAndStoresExampleDriver`.

src/main/java/io/confluent/examples/streams/GlobalStoresExample.java
```diff
@@ -0,0 +1,242 @@
...
+    public static void main(final String[] args) {
+        final String bootstrapServers = args.length > 0 ? args[0] : "localhost:9092";
+        final String schemaRegistryUrl = args.length > 1 ? args[1] : "http://localhost:8081";
+        final KafkaStreams
+                streams =
+                createStreams(bootstrapServers, schemaRegistryUrl, "/tmp/kafka-streams-global-stores");
+        // Always (and unconditionally) clean local state prior to starting the processing topology.
+        // We opt for this unconditional call here because this will make it easier for you to play around with the example
+        // when resetting the application for doing a re-run (via the Application Reset Tool,
+        // http://docs.confluent.io/current/streams/developer-guide.html#application-reset-tool).
```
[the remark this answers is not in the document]
adrien-ben: Updated all instances in a separate commit

adrien-ben:
> Thanks for the PR @adrien-ben!
> 
> About test code duplication: seems reasonable to me to avoid the duplicated code, as we literally test the same input data and expect the same result.

So, you think I should add some helper class in the tests containing the record injection and assertions ?

mjsax:
> So, you think I should add some helper class in the tests containing the record injection and assertions ?

I was thinking about an abstract parent class that unifies the shares logic, and has a method (something like) `abstract Topology topology()` that the two actual tests implement.

adrien-ben:
Done

mjsax (approved):
Thanks for the PR! Great work!

PR #375, opened 2021-02-05
MINOR: use isolated state.dir for each KafkaStreams client

mjsax:
With https://github.com/apache/kafka/pull/9978 we start to lock the state directory and disallow to share it across multiple `KafkaStreams` instances.

### DIFF
--- a/src/main/java/io/confluent/examples/streams/microservices/OrdersService.java
+++ b/src/main/java/io/confluent/examples/streams/microservices/OrdersService.java
@@ -325,6 +325,7 @@ public class OrdersService implements Service {
     jettyServer = startJetty(port, this);
     port = jettyServer.getURI().getPort(); // update port, in case port was zero
     producer = startProducer(bootstrapServers, ORDERS, defaultConfig);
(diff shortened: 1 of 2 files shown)


guozhangwang (approved):

ableegoldman:
@mjsax looks like `OrdersServiceTest.shouldGetOrderByIdWhenOnDifferentHost` still failed. Checking on the logs, I see
```[2021-02-05 23:06:42,784] WARN [main] Using an OS temp directory in the state.dir property can cause failures with writing the checkpoint file due to the fact that this directory can be cleared by the OS. Resolved state.dir: [/tmp/kafka-streams]```
...so it seems like it's still using the same state dir somewhere

mjsax:
Updated this PR. Might be useful to actually set the state dir in the config instead of ignoring the parameter....

ableegoldman:
> Might be useful to actually set the state dir in the config instead of ignoring the parameter....
Hahahaha oh no

ableegoldman (approved):
K this LGTM, assuming the test passes this time

mjsax:
Nope... Will dig into it next week...

ableegoldman:
Thank you 🙏

PR #387, opened 2021-04-10
MINOR: fix compiler warnings

mjsax:
Added compiler warnings about deprecated API usage. Plus a few minor cleanups.


src/main/java/io/confluent/examples/streams/microservices/domain/Schemas.java
```diff
@@ -93,8 +92,6 @@ private static void createTopics() {
       map.put(BASIC_AUTH_CREDENTIALS_SOURCE, config.getProperty(BASIC_AUTH_CREDENTIALS_SOURCE));
     if (config.containsKey(USER_INFO_CONFIG))
       map.put(USER_INFO_CONFIG, config.getProperty(USER_INFO_CONFIG));
-    else if (config.containsKey(SCHEMA_REGISTRY_USER_INFO_CONFIG))
-      map.put(USER_INFO_CONFIG, config.getProperty(SCHEMA_REGISTRY_USER_INFO_CONFIG));
```
vvcephei, line 97: What's up with this change?

vvcephei (approved):
Thanks for going through the rest of the warnings, @mjsax !

src/main/java/io/confluent/examples/streams/microservices/domain/Schemas.java
vvcephei: Ah, I see. That's an static import from SR, which is deprecated. Nevermind :)

PR #389, opened 2021-05-04
MINOR: Update test to match new opererator implementation

mjsax:
Stream-Stream left-/outer-joins got improved via
https://issues.apache.org/jira/browse/KAFKA-10847 to avoid spurious
left/outer join results.

This PR updates the stream-stream join test to match the new join
implemenation.


spena (approved):
lgtm

lct45 (approved):
Thanks for doing this so quickly @mjsax. Looks good to me

PR #391, opened 2021-06-11
MINOR: bump timeout to stabilize test

mjsax:
Running this test locally, if failed reliably with the 30sec timeout and the issue went await with the 60sec timeout.


cprasad1 (approved):
LGTM

PR #394, opened 2021-07-07
Migrate off of deprecated SchemaRegistry zk config, use bootstrap servers config instead

ableegoldman:
In [#1936](https://github.com/confluentinc/schema-registry/commit/06738ad7ce75cb1105a65d8356938751d7d68273#diff-685a652ce1af196ad342417568034ad869e6f3875a886566652064935fc64bdeL645-L655), SchemaRegistry started to enforce that `KAFKASTORE_BOOTSTRAP_SERVERS_CONFIG` be set, rather than allowing the app to use either that or `KAFKASTORE_CONNECTION_URL_CONFIG` which turned out to. be a deprecated zookeeper-specific config that we apparently still used.

### DIFF
--- a/src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
+++ b/src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
@@ -169,17 +169,6 @@ public class EmbeddedSingleNodeKafkaCluster extends ExternalResource {
   }
 
   /**
-   * This cluster's ZK connection string aka `zookeeper.connect` in `hostnameOrIp:port` format.
-   * Example: `127.0.0.1:2181`.
-   * <p>
-   * You can use this to e.g. tell Kafka consumers (old consumer API) how to connect to this
-   * cluster.
-   */
-  public String zookeeperConnect() {
-    return zookeeper.connectString();
-  }
-
-  /**
    * The "schema.registry.url" setting of the schema registry instance.
    */
   public String schemaRegistryUrl() {
--- a/src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
+++ b/src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
@@ -43,7 +43,7 @@ import java.util.Set;
 public class EmbeddedSingleNodeKafkaCluster extends ExternalResource {
 
   private static final Logger log = LoggerFactory.getLogger(EmbeddedSingleNodeKafkaCluster.class);
-  private static final int DEFAULT_BROKER_PORT = 0; // 0 results in a random port being selected
+  private static final int DEFAULT_BROKER_PORT = 1234; // pick a random port
   private static final String KAFKA_SCHEMAS_TOPIC = "_schemas";
   private static final String AVRO_COMPATIBILITY_TYPE = CompatibilityLevel.NONE.name;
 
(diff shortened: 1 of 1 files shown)


src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
```diff
@@ -96,8 +96,9 @@ public void start() throws Exception {
     schemaRegistryProps.put(SchemaRegistryConfig.KAFKASTORE_TIMEOUT_CONFIG, KAFKASTORE_OPERATION_TIMEOUT_MS);
     schemaRegistryProps.put(SchemaRegistryConfig.DEBUG_CONFIG, KAFKASTORE_DEBUG);
     schemaRegistryProps.put(SchemaRegistryConfig.KAFKASTORE_INIT_TIMEOUT_CONFIG, KAFKASTORE_INIT_TIMEOUT);
+    schemaRegistryProps.put(SchemaRegistryConfig.KAFKASTORE_BOOTSTRAP_SERVERS_CONFIG, effectiveBrokerConfig.getProperty(KafkaConfig.ListenersProp()));
 
-    schemaRegistry = new RestApp(0, zookeeperConnect(), KAFKA_SCHEMAS_TOPIC, AVRO_COMPATIBILITY_TYPE, schemaRegistryProps);
+    schemaRegistry = new RestApp(0, null, KAFKA_SCHEMAS_TOPIC, AVRO_COMPATIBILITY_TYPE, schemaRegistryProps);
```
ableegoldman, line 101: We never actually explicitly set the deprecated zk config anywhere, turned out it was implicitly being set inside `RestApp` when you pass in the connection here. So we can (and should) just remove this

src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
```diff
@@ -43,7 +43,7 @@
 public class EmbeddedSingleNodeKafkaCluster extends ExternalResource {
 
   private static final Logger log = LoggerFactory.getLogger(EmbeddedSingleNodeKafkaCluster.class);
-  private static final int DEFAULT_BROKER_PORT = 0; // 0 results in a random port being selected
+  private static final int DEFAULT_BROKER_PORT = 1234; // pick a random port
```
ableegoldman, line 46: I had some trouble getting this to run locally without setting an explicit port, rather than the current value of `DEFAULT_BROKER_PORT` which is `0`. According to the comment on that line this results in a random port being selected, but in my experience it was actually just selecting the literal `0`. So, I'm picking a random port here. If anyone knows how the "0 --> random port" thing is supposed to work, and why it now isn't, I can look into fixing that up. But as it is I don't really know where to go from here and thus we're just going with `1234`

rayokota (approved):
Thanks @ableegoldman , LGTM

ableegoldman:
Merging now since this is causing pretty much all tests here to fail. If anyone from @confluentinc/streams has any further questions or feedback I can address it in a followup PR

src/test/java/io/confluent/examples/streams/kafka/EmbeddedSingleNodeKafkaCluster.java
vvcephei: That's probably fine. I don't know how/whether the randomization thing is supposed to work either.

The only concern would be the likelihood of colliding with another process already running on port 1234, which would result in the test failing to bind the port.

From the looks of it, port 1234 is used by VLC and maybe a Monopoly game server? (https://duckduckgo.com/?t=ffab&q=port+1234), so I think we're fine.

PR #395, opened 2021-07-08
fix: update lambda integration test with new null serde changes

lct45:
We recently [changed the default serde to be null](https://github.com/apache/kafka/pull/10813) and therefore need to update this test to make sure it properly sets a key serde

### DIFF
--- a/src/test/java/io/confluent/examples/streams/FanoutLambdaIntegrationTest.java
+++ b/src/test/java/io/confluent/examples/streams/FanoutLambdaIntegrationTest.java
@@ -72,6 +72,7 @@ public class FanoutLambdaIntegrationTest {
     streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "fanout-lambda-integration-test");
     streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
     streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
+    streamsConfiguration.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.ByteArray().getClass().getName());
 
     final String inputTopicA = "A";
     final String outputTopicB = "B";


src/test/java/io/confluent/examples/streams/FanoutLambdaIntegrationTest.java
```diff
@@ -72,6 +72,7 @@ public void shouldFanoutTheInput() {
     streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "fanout-lambda-integration-test");
     streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
     streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
+    streamsConfiguration.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.ByteArray().getClass().getName());
```
ableegoldman, line 75: Weird, this test really did use the ByteArraySerde. Tbh maybe we should just convert it to a more human readable type, like a String or Integer or something? Not sure how much work that would be, but if it's just ~10 LOC then let's just clean up the test

If it would be a big overhaul then I'm fine with just explicitly setting it to the ByteArraySerde, it just seems bizarre to actually be using this, especially in an example

lct45: I just switched it to be string and it didn't complain 🤔  I think because it's using the topology test driver we need the config but if you look further down in the test it looks like we're passing in `new IntegrationTestUtils.NothingSerde<>()`

ableegoldman: I guess the StringSerde is also not too picky, since it's just an arbitrary length of bytes same as the byte array. It would be gibberish though, the test is clearing _intending_ to use the ByteArraySerde since the types are things like `KStream<byte[], String>`

If we really don't care at all about the key, and maybe we don't, then I would just keep it as a ByteArraySerde. Otherwise it's confusing since users see that it's a StringSerde in the configs, but the KStream type shows it as a `byte[]`

src/test/java/io/confluent/examples/streams/FanoutLambdaIntegrationTest.java
```diff
@@ -72,6 +72,7 @@ public void shouldFanoutTheInput() {
     streamsConfiguration.put(StreamsConfig.APPLICATION_ID_CONFIG, "fanout-lambda-integration-test");
     streamsConfiguration.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy config");
     streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
+    streamsConfiguration.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.String().getClass().getName());
```
ableegoldman, line 75: @lct45  Ok sorry for the back and forth, not sure if you saw my last comment but let's actually just leave it as a ByteArraySerde for now and I can merge it

src/test/java/io/confluent/examples/streams/FanoutLambdaIntegrationTest.java
lct45: It works with a string serde too.. It seems like because it uses the topology test driver we set the serdes earlier? So it requires the config to be set when we initialize the topology test driver

src/test/java/io/confluent/examples/streams/FanoutLambdaIntegrationTest.java
lct45: Ah sorry @ableegoldman just responded - it working with byte array and string seems slightly concerning to me (lmk what you think tho). I can't dive into this today but can tomorrow

ableegoldman: I'm not surprised it would work with both, an IntegerSerde for example can tell when the data type is not right because it has to always be 4 bytes, whereas a StringSerde could happily interpret any variable length series of bytes as a string (it's just that the string may be complete nonsense).

So let's just set it to be a ByteArraySerde and leave it at that. Sorry for making this confusing 😅

ableegoldman: If you get the chance to change it back into a ByteArraySerde today then I can merge it, otherwise if you get to it tomorrow then can you just merge it since I'll be on a plane?

lct45: No worries, thanks for diving into this. just updated

ableegoldman (approved):
LGTM, thanks for the fix!

PR #400, opened 2021-08-09
Correct internal API to use string instead of boolean

lct45:
saw a bunch of failures here: https://confluentinc.atlassian.net/browse/KSQL-6263
recent logs indicate that there was a parameter mismatch:
`[2021-08-09T13:45:35.897Z] [ERROR] /home/jenkins/workspace/nc_kafka-streams-examples_master/src/main/scala/io/confluent/examples/streams/algebird/CMSStoreChangeLogger.scala:36: error: type mismatch;
[2021-08-09T13:45:35.897Z] [ERROR]  found   : Boolean(false)
[2021-08-09T13:45:35.897Z] [ERROR]  required: String
[2021-08-09T13:45:35.897Z] [ERROR]   private val topic = ProcessorStateManager.storeChangelogTopic(context.applicationId, storeName, false)`

the current internal api takes `public static String storeChangelogTopic(final String applicationId, final String storeName, final String namedTopology)`

PR #401, opened 2021-08-09
MINOR: fix compile error

mjsax:
We use internal APIs that got changed upstream. Need to update the code
accordingly.

### DIFF
--- a/src/main/scala/io/confluent/examples/streams/algebird/CMSStoreChangeLogger.scala
+++ b/src/main/scala/io/confluent/examples/streams/algebird/CMSStoreChangeLogger.scala
@@ -33,7 +33,7 @@ class CMSStoreChangeLogger[K, V](val storeName: String,
                                  val partition: Int,
                                  val serialization: StateSerdes[K, V]) {
 
-  private val topic = ProcessorStateManager.storeChangelogTopic(context.applicationId, storeName, false)
+  private val topic = ProcessorStateManager.storeChangelogTopic(context.applicationId, storeName, null)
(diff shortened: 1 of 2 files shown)


lct45:
I also have a PR here: https://github.com/confluentinc/kafka-streams-examples/pull/400, I think we need to add the fix to `InventoryServiceTest`

src/main/scala/io/confluent/examples/streams/algebird/CMSStoreChangeLogger.scala
```diff
@@ -33,7 +33,7 @@ class CMSStoreChangeLogger[K, V](val storeName: String,
                                  val partition: Int,
                                  val serialization: StateSerdes[K, V]) {
 
-  private val topic = ProcessorStateManager.storeChangelogTopic(context.applicationId, storeName, false)
+  private val topic = ProcessorStateManager.storeChangelogTopic(context.applicationId, storeName, null)
```
lct45, line 36: any reason why we're passing `null` instead of `context.taskId().namedTopology()`?

wcarlson5: I don't think that this should be using the `namedTopology` stuff yet. We have not really finished that work 100% yet

mjsax: I am not familiar with the details of the upstream work. Did not even know that `context.taskId().namedTopology()` exists... Let me try this.

mjsax:
Closing this PR, as there is already https://github.com/confluentinc/kafka-streams-examples/pull/400

PR #402, opened 2021-08-10
MINOR: fix compile error

mjsax:
The changed API is public, but it's not release yet, thus there is no upstream deprecation necessary, but the KIP was changed.


wcarlson5 (approved):
Thanks

PR #403, opened 2021-08-15
MINOR: Fix sum-lambda example

### LINKED ISSUE #326: Cast error
Class: SumLambdaExample
line: final KStream<Void, Integer> input = builder.stream(NUMBERS_TOPIC);

this declaration makes application crash due to a ClassCastException when SumLambdaExampleDriver sends data. It tries to convert Integer (the type of the key) to type Void.

mjsax:
Closes #326


bbejeck (approved):
LGTM!

Issue #179, opened 2018-10-17
Question: Adding example for reordering out of order messages


1123:
I recently came across this question: https://groups.google.com/forum/#!topic/confluent-platform/5NrqRz-GITk . Would it make sense to contribute an example stateful transformer in this repository? I would offer to contribute this.

miguno:
@1123 : If you'd like to contribute such a stateful example, then this would be great!

sshcherbakov:
I sketched up an example for reordering using stateful transformer.
Would be glad to discuss.

bbejeck:
Hi, @sshcherbakov would you consider contributing this to https://github.com/confluentinc/kafka-tutorials? 
I can help you with the tutorial writing process if you want.

mjsax:
PR: https://github.com/confluentinc/kafka-streams-examples/pull/411

sshcherbakov:
Thank you @bbejeck and apologies for delay in response. 
I checked with my employer and it seems that I may contribute to https://github.com/confluentinc/kafka-tutorials

Thanks @mjsax for the thorough review, I'll correct. Btw. there might be an issue with the not fixed time zone. I'll try to improve that as well.

mjsax:
Should we still merge #441 or move directly to a kafka-tutorials contribution? No harm to merge, but the examples repo is superseded with the tutorial so not sure how much sense it makes to add to ti?

bbejeck:
If @sshcherbakov is willing to do a tutorial then I'd say move it to a kafka-tutorials contribution.  I'll leave it to @sshcherbakov if he wants to close #411.

bbejeck:
@sshcherbakov if you have any questions or need any help feel free to reach out to me.

sshcherbakov:
I hope to have addressed all comments and updated PR.
Also fixed time zone to UTC and made a couple of minor simplifications.

PR #411, opened 2021-11-25
Stream reorder example. Issue #179

sshcherbakov:
Example of reordering out of order messages using a stateful transformer.

### DIFF
--- a/src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
+++ b/src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
@@ -1,0 +1,239 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.confluent.examples.streams;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.serialization.LongDeserializer;
+import org.apache.kafka.common.serialization.LongSerializer;
+import org.apache.kafka.common.serialization.Serdes;
+import org.apache.kafka.streams.*;
+import org.apache.kafka.streams.kstream.KStream;
+import org.apache.kafka.streams.kstream.Named;
+import org.apache.kafka.streams.kstream.Transformer;
+import org.apache.kafka.streams.processor.ProcessorContext;
+import org.apache.kafka.streams.processor.PunctuationType;
+import org.apache.kafka.streams.processor.TimestampExtractor;
+import org.apache.kafka.streams.state.KeyValueIterator;
+import org.apache.kafka.streams.state.KeyValueStore;
+import org.apache.kafka.streams.state.StoreBuilder;
+import org.apache.kafka.streams.state.Stores;
+import org.junit.Test;
+
+import java.text.ParseException;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Properties;
+
+import static java.time.temporal.ChronoUnit.HOURS;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * End-to-end integration test that demonstrates how to reorder the stream of incoming messages
+ * by the timestamp embedded in the message payload.
+ * <p>
+ * Makes sense only on per partition basis.
+ * <p>
+ * Reordering occurs within the time windows defined by the {@code grace} constructor parameter.
+ *
+ * Note: This example uses lambda expressions and thus works with Java 8+ only.
+ */
+public class ReorderIntegrationTest {
+
+  public static class ReorderTransformer<K, V>
+      implements Transformer<K, V, KeyValue<K, V>> {
+
+    public interface StoreKeyGenerator<K, V> {
+      K getStoreKey(K key, V val);
+    }
+
+    private final String storeName;
+    private final Duration grace;
+    private KeyValueStore<K, V> reorderStore;
+    private ProcessorContext context;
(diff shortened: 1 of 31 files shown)


src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
```diff
@@ -0,0 +1,248 @@
...
+     * Outputs downstream accumulated records sorted by their timestamp.
+     *
+     * 1) read the store using a ranged fetch from 0 to timestamp - 60'000 (=1 minute)
+     * 2) send the fetched messages in order using context.forward() and deletes
+     * them from the store
+     *
+     * @param timestamp – stream time of the punctuate function call
+     */
+    void punctuate(final long timestamp) {
+      try(KeyValueIterator<K, V> it = reorderStore.all()) {
```
bbejeck, line 128: The comments state to use `range(from, to)`, but the code uses `all()` since the code deletes everything in the store after forwarding, I think it will continue to work as-is, but you should update the comments.

bbejeck (commented):
Thanks for the PR @sshcherbakov looks good overall, I just have a couple of comments.

src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
```diff
@@ -0,0 +1,248 @@
...
+ * <p>
+ * Makes sense only on per partition basis.
+ * <p>
+ * Reordering occurs within time windows defined by the
+ *
+ * Note: This example uses lambda expressions and thus works with Java 8+ only.
+ */
+public class ReorderIntegrationTest {
+
+  public static class ReorderTransfomer<K, V>
```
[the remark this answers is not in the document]
sshcherbakov: Fixed

src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
```diff
@@ -0,0 +1,248 @@
...
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * End-to-end integration test that demonstrates how to reorder the stream of incoming messages
+ * by the timestamp embedded in the message payload.
+ * <p>
+ * Makes sense only on per partition basis.
+ * <p>
+ * Reordering occurs within time windows defined by the
```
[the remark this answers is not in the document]
sshcherbakov: Amended

src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
```diff
@@ -0,0 +1,248 @@
...
+     * as the key and the message-value as the value, this also provides
+     * de-duplication.
+     *
+     * @param key
+     * @param value
+     * @return
+     */
+    @Override
+    public KeyValue<K, V> transform(final K key, final V value) {
+      final long ts = timestampExtractor.getTimestamp(value);
```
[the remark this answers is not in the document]
sshcherbakov: The Kafka Streams TimestampExtractor could have been reused to illustrate pulling the timestamp from the payload, but the method signature is not convenient. Instead of duplicating TimestampExtractor interface with convenient signature removed removed this piece to reduce confusion.

src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
sshcherbakov: Updated comments

sshcherbakov:
Rebased PR on top of the current master branch. I [cannot build master](https://github.com/confluentinc/common/issues/495) but I checked the test with the latest release branch `7.3.1-post` and it still passes.

src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
```diff
@@ -0,0 +1,239 @@
...
+     * 2) send the fetched messages in order using context.forward() and delete
+     * them from the store
+     *
+     * @param timestamp – stream time of the punctuate function call
+     */
+    void punctuate(final long timestamp) {
+      try(KeyValueIterator<K, V> it = reorderStore.all()) {
+        while (it.hasNext()) {
+          final KeyValue<K, V> kv = it.next();
+          context.forward(kv.key, kv.value);
```
singhbaljit, line 124: I think this implementation is flawed. Ordering only makes sense per partition/key. 

Say incoming records have key `k1`, and there are two messages with timestamps `t0`, and `t1`. The `StoreKeyGenerator` will map that into some combination of the original key and the record timestamp, say something simple like `k1_t0` and `k1_t1`. However, when you forward these message with the "composite" store key, they'll likely end up in two different partitions. Hence, although you sorted them, any downstream processor/consumer has no guarantee in the order after the re-partition. The correct thing to do would be to preserve the original key, and use that to forward the message.

src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
```diff
@@ -0,0 +1,239 @@
...
+     * @param value
+     * @return
+     */
+    @Override
+    public KeyValue<K, V> transform(final K key, final V value) {
+      // Keys need to contain and be sortable by time
+      final K storeKey = storeKeyGenerator.getStoreKey(key, value);
+      final V storeValue = reorderStore.get(storeKey);
+      if(storeValue == null) {
+        reorderStore.put(storeKey, value);
```
singhbaljit, line 103: putIfAbsent?

sshcherbakov: Do you mean pitIfAbsent() is equivalent or it has different logic?
In its description: "_Update the value associated with this key, unless a value is already associated with the key._"
Before I start interpreting that and guessing whether it will only put my value where there is no key or even when key is there but the value is null, I write my intent explicitly through additional "if" line here.

src/test/java/io/confluent/examples/streams/ReorderIntegrationTest.java
sshcherbakov: It is an example. It can be extended with either additional store for composite to original key mapping or a second interface that would allow extracting original key from the composite one before passing downstream.

bbejeck:
We are currently starting to migrate the examples in this repository to https://github.com/confluentinc/tutorials,  so it didn't make sense to merge this here just to move it later.  There is a [PR now under review](https://github.com/confluentinc/tutorials/pull/53). As a result, I'm closing this PR now.

\cc @mjsax , @sshcherbakov

bbejeck:
Thank you @sshcherbakov for the awesome contribution!

PR #427, opened 2022-06-21
MINOR: fix master

mjsax:
Upstream change broke the code:
https://github.com/apache/kafka/commit/a6c5a74fdbdce9a992b47706913c920902cda28c#diff-84b3ac8682467c5c0fe024978e60e02c6f4583d3c72171ff20c41a419ca04b84)

### DIFF
--- a/src/main/scala/io/confluent/examples/streams/algebird/CMSStore.scala
+++ b/src/main/scala/io/confluent/examples/streams/algebird/CMSStore.scala
@@ -194,7 +194,7 @@ class CMSStore[T: CMSHasher](override val name: String,
       name,
       Serdes.Integer(),
       TopCMSSerde[T])
-    changeLogger = new CMSStoreChangeLogger[Integer, TopCMS[T]](name, context, serdes)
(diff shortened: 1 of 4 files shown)


src/main/scala/io/confluent/examples/streams/algebird/CMSStore.scala
```diff
@@ -194,7 +194,7 @@ class CMSStore[T: CMSHasher](override val name: String,
       name,
       Serdes.Integer(),
       TopCMSSerde[T])
-    changeLogger = new CMSStoreChangeLogger[Integer, TopCMS[T]](name, context, serdes)
+    changeLogger = new CMSStoreChangeLogger[Integer, TopCMS[T]](name, context, serdes, name)
```
cadonna, line 197: Are you sure it is fine to use the same name for the state store and the processor node ID?

mjsax: I actually set the processor-name so it should match up now.

PR #471, opened 2024-02-02
chore: disable semaphore

lucasbru:
Finishing the semaphore migration is blocked, see https://confluentinc.atlassian.net/browse/KSTREAMS-5779.

Meanwhile, the semaphore job is failing and creating lots of testbreak ticket noise, so I'm trying to disable it.

### DIFF
--- a/service.yml
+++ b/service.yml
@@ -13,6 +13,8 @@ semaphore:
   maven_skip_deploy: true
   build_arm: true
   nano_version: true
+  branches: []
+  triggers: []
 
 # Properties not ported from Jenkinsfile:
 #


cadonna (approved):

service.yml
```diff
@@ -13,10 +13,12 @@ semaphore:
...
+  triggers: []
 
 # Properties not ported from Jenkinsfile:
 #
 # upstreamProjects = 'confluentinc/rest-utils' --- used "downstream_repos in rest-utils
 # cpImages = true
 # osTypes = ['ubi8']
-# slackChannel = 'ksqldb-warn'
\ No newline at end of file
+# slackChannel = 'ksqldb-warn'
```
mjsax, line 24: @lucasbru -- while semaphor is not running yet, this slack channel seems to be outdate? -- We we re-enable is, can we also update the slack channel?

PR #478, opened 2024-03-06
MINOR: fix test after upstream change

mjsax:
Upstream fix: https://github.com/apache/kafka/pull/14426

### DIFF
--- a/src/test/java/io/confluent/examples/streams/StreamToStreamJoinIntegrationTest.java
+++ b/src/test/java/io/confluent/examples/streams/StreamToStreamJoinIntegrationTest.java
@@ -134,6 +134,7 @@ public class StreamToStreamJoinIntegrationTest {
       impressionInput.pipeKeyValueList(inputAdImpressions);
       clickInput.pipeKeyValueList(inputAdClicks);
       // Flush result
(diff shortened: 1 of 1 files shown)


src/test/java/io/confluent/examples/streams/StreamToStreamJoinIntegrationTest.java
```diff
@@ -134,6 +134,7 @@ public void shouldJoinTwoStreams() {
       impressionInput.pipeKeyValueList(inputAdImpressions);
       clickInput.pipeKeyValueList(inputAdClicks);
       // Flush result
+      topologyTestDriver.advanceWallClockTime(Duration.ofSeconds(1L));
```
mjsax, line 137: It's a little unclear why this was not necessary to begin with... w/o advancing wall-clock time, we might not flush due to wall-clock time base "flush throttling".

This was also notices on the upstream PR (https://github.com/apache/kafka/pull/14426#issuecomment-1976647985) so it actually seems the PR did change this behavior a little bit (unclear what the exact root cause is -- nobody did investigate further, and seems it's not worth to spend time on?)

wcarlson5 (approved):

PR #488, opened 2024-04-18
MINOR: disable tag builds

lucasbru:
As per comments on https://confluentinc.atlassian.net/browse/KSTREAMS-5737, we should just disable tag builds


cadonna (approved):
LGTM

PR #490, opened 2024-04-18
MINOR: Fix config names

lucasbru:
Commit 61baa7ac6bb871797197d9289a848a0d4f587ced in Kafka moved around the internal fields for these public configs.

We should not depend on the internal fields, otherwise we will keep breaking here. Just use the public API.


cadonna (approved):
LGTM

PR #496, opened 2024-04-22
Sureshpidikiti/epoch converter example

sureshpidikiti:
Adding a new example to convert the Unix timestamp seconds in each Kafka record to a timestamp format using JAVA SDF and Streams mapValues.


mjsax:
Thanks for the PR @sureshpidikiti -- we don't add new examples to this repos though, as we migrated the content to https://github.com/confluentinc/tutorials \cc @bbejeck for visibility.

Closing this PR.

PR #523, opened 2024-10-08
MINOR: fix build broken references and params

lianetm:
Fixing build. 

1 - Remove unsupported method that was overriding one that was recently removed with upstream change https://github.com/apache/kafka/pull/16906/files#diff-7fe11ab5459b6c6e35ee9699639025e1ce8814b8b603596f69f6e45a011e1675

2 - fix wrong param that was causing abrupt failure in ApplicationResetIntegrationTest (test crashed). Mixed in the logs there was:

      bootstrap-servers is not a recognized option
      Option (* = required) Description
      --bootstrap-server <String: server to The server(s) to connect to....

All test pass locally after these fixes.


philipnee (approved):
thanks

lucasbru (approved):
LGTM, thanks!

PR #527, opened 2024-12-12
7.1.x servicebot update

hk10111:
Automated update by servicebot


mjsax:
@hk10111 -- Why are we pathing 7.1.x branch?

Should we not just only update `master` and bump Java version to 17 there? Sound much simpler than messing with older branches?

hk10111:
@mjsax it would be better to keep the pipelines consistent across branches rather than making the changes only in master

mjsax (approved):

PR #535, opened 2025-03-27
Pinned to jetty version 9, removed monitoring-interceptors deprecated dependency

apoorvmittal10:
The commmit [here](https://github.com/confluentinc/common/commit/60322b8531dd93a1afc765080a43c3cb13392223) has moved the jetty.version to 12.0.16, but `jetty-servlet`(https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-servlet) has been moved to `jetty-ee10-servlet` in version 12 hence the current pom artifacts can't be find.

This PR pins the jetty-servlet version to fix the builds. We should aim for migration to version 12 of jetty servlet soon after. Also PR removes the deprecated dependecy on `monitoring-interceptors`.


mjsax (approved):

bbejeck (approved):

bbejeck (approved):
Thanks for the PR @apoorvmittal10 !

PR #583, opened 2025-07-16
DP-16708: Switch to using confluent-devtools

nlou9:
As part of the [depot migration task](https://confluentinc.atlassian.net/browse/DP-16108), replace release-tools with confluent-devtools. Remove all the installation of release-tools. We don't need to install release-tools on CI since it is preinstalled on CI AMI images. 

One Pager: https://confluentinc.atlassian.net/wiki/spaces/TOOLS/pages/4020831896/Depot+-+Python+Walled+Garden+Repository


aliehsaeedii (approved):
LGTM.

PR #666, opened 2026-08-11
Replace removed Apache Kafka test helpers with local copies

aliehsaeedii:
`org.apache.kafka.test.TestUtils` and `TestCondition` moved to test fixtures in Apache Kafka [PR #22201](https://github.com/apache/kafka/pull/22201), so they're no longer in the published kafka-clients test artifact this project resolves — test compilation currently fails.                                                                                                                                                                                               
                                                                                                                                                                                                                 
This PR adds local ports of the only helpers used (`tempDirectory()`, `waitForCondition`, `TestCondition`) under `io.confluent.examples.streams.utils`, keeping upstream timings. Same class name, so call sites are unchanged 
  — the other 34 files change one import line each.


bbejeck (approved):

src/test/java/io/confluent/examples/streams/utils/TestCondition.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package io.confluent.examples.streams.utils;
+
+/**
+ * Interface to wrap actions that are required to wait until a condition is met for testing
+ * purposes. Note that this is not intended to do any assertions.
+ *
+ * <p>Adapted from {@code org.apache.kafka.test.TestCondition} in Apache Kafka. That class moved
+ * from the {@code clients} test source set to test fixtures in Apache Kafka PR #22201, and is
```
mjsax, line 24: But this problem should go away, after AK 4.4 was released I assume? After the release, there should be a new jar and we could update our build to depend on it, so we don't have to duplicate code?

src/test/java/io/confluent/examples/streams/utils/TestUtils.java
```diff
@@ -0,0 +1,207 @@
...
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.function.Supplier;
+
+/**
+ * Test helpers adapted from {@code org.apache.kafka.test.TestUtils} in Apache Kafka.
+ *
+ * <p>Those helpers moved from the {@code clients} test source set to test fixtures in Apache
+ * Kafka PR #22201, so they are no longer present in the published {@code kafka-clients}
```
mjsax, line 32: Same

src/test/java/io/confluent/examples/streams/utils/TestCondition.java
aliehsaeedii: You mean we should use the fixtures then? I just thought better not to rely on AK for those helpers any more.

PR #667, opened 2026-08-11
Run build and test blocks on master, not only on pull requests

aliehsaeedii:
Both build blocks are gated `when: "pull_request =~ '.*'"`, so pushes to master skip every block — nothing is compiled or tested, and a green pipeline carries no signal.

Gates both build blocks on master too, and adds `managed_sections` omitting `blocks` to `service.yml` so ServiceBot preserves that gating (the `cp-dockerfile` default includes `blocks`, so otherwise it is regenerated away).

The `or` in the `when:` expression evaluates correctly for the PR case — the build blocks ran on this PR rather than being skipped. The master case can only be confirmed by the first master build after merge, and `managed_sections` only takes effect on the next ServiceBot run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -117,7 +117,7 @@ blocks:
   - name: Build, Test, & Scan AMD
     dependencies: ["Validation"]
     run:
-      when: "pull_request =~ '.*'"
+      when: "pull_request =~ '.*' or branch = 'master'"
     task:
       jobs:
         - name: Build, Test, & Scan ubi9
@@ -146,7 +146,7 @@ blocks:
   - name: Build & Test ARM
     dependencies: ["Validation"]
     run:
-      when: "pull_request =~ '.*'"
+      when: "pull_request =~ '.*' or branch = 'master'"
     task:
       agent:
         machine:
--- a/service.yml
+++ b/service.yml
@@ -12,6 +12,18 @@ codeowners:
 semaphore:
   enable: true
   pipeline_type: cp-dockerfile
+  # 'blocks' is deliberately omitted so ServiceBot does not regenerate it. The build blocks in
+  # .semaphore/semaphore.yml are gated to also run on master; without this, that gating is reset
+  # on the next ServiceBot regeneration and master builds nothing again.
+  managed_sections:
+    - version
+    - name
+    - agent
+    - fail_fast
+    - execution_time_limit
+    - queue
+    - global_job_config
+    - after_pipeline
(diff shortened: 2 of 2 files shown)


.semaphore/semaphore.yml
```diff
@@ -115,9 +115,9 @@ blocks:
           commands:
             - ci-sem-pint -c
   - name: Build, Test, & Scan AMD
-    dependencies: ["Validation"]
+    dependencies: []
     run:
-      when: "pull_request =~ '.*'"
+      when: "pull_request =~ '.*' or branch = 'master'"
```
aliehsaeedii: Re: @copilot — release branches (`1.2.x`, `1.2.3`, `-rc1`, …) will still skip this block unless they're PRs, so extend the condition with the release-branch regexes from `.semaphore/cp_dockerfile_build.yml:157`.

The premise isn't quite right: the PR description only calls out **master** ("pushes to master skip every block"), not release branches — restoring the master build signal is the deliberate, intended scope here. Building on release-branch pushes is a separate design decision (release branches already have their own gating in `cp_dockerfile_build.yml`), so I'm keeping this change master-only. Happy to add the release-branch regexes in a follow-up if we decide those pushes should build too.

.semaphore/semaphore.yml
```diff
@@ -115,9 +115,9 @@ blocks:
...
     task:
       jobs:
         - name: Build, Test, & Scan ubi9
```
aliehsaeedii: This job also runs `cve-scan` and pushes a dev image, so every master commit now does that too, not just build and test. Intended, or do you only want the compile/test signal?

service.yml
```diff
@@ -12,6 +12,18 @@ codeowners:
 semaphore:
   enable: true
   pipeline_type: cp-dockerfile
+  # 'blocks' is deliberately omitted so ServiceBot does not regenerate it. The build blocks in
+  # .semaphore/semaphore.yml are gated to also run on master; without this, that gating is reset
+  # on the next ServiceBot regeneration and master builds nothing again.
+  managed_sections:
```
aliehsaeedii: Leaving `blocks` out freezes the whole section, not just the `when:` lines — ServiceBot copies the existing `blocks` over the generated one, so future template changes to the build/scan/push steps stop landing here, and `os_types`, `build_arm` and `maven_phase` in this file no longer affect the pipeline. Worth asking Foundations for a narrower knob before taking that on permanently.

